PR #6, opened 2019-12-26
Resolves #2: basic unit/element model

### LINKED ISSUE #2: Create a basic model for units
We need a model to represent the hierarchy of military units to some degree.
The rough idea so far is something like:
There are `Elements` and `Units`. There might be `Units` of types
- Soldier
- Helicopter
- APC
And they make up different kind of `Elements` while `Elements` themselves can again make up bigger `Elements`. For example
- 2 Soldiers (`Unit`) are a Buddyteam (`Element`)
- 2 Buddyteams are a Fireteam (`Element`)
- 2 Fireteams are a Squad (`Element`)
- etc...

For the beginning maybe a simple structure of `Platoon` consisting of `Soldier`s might be enough.

Loomie:


src/main/kotlin/de/gleex/pltcmd/model/elements/ArmyUnitHierarchy.kt
```diff
@@ -0,0 +1,26 @@
...
+	/** Formation, strength: 3,000 to 5,000  */
+	Brigade(3, 6),
+	/** Unit, strength: 300 to 1,000  */
+	Battalion(2, 6),
+	/** Subunit, strength: 80 to 250  */
+	Company(2, 8),
+	/** Sub-subunit, strength: 26 to 55  */
+	Platoon(2, 5),
+	/** strength: 12 to 24  */
+	Section(1, 4),
```
Baret: To keep it simple I would rather like to remove section and have a platoon directly consist of squads.

src/main/kotlin/de/gleex/pltcmd/model/elements/ArmyUnitHierarchy.kt
```diff
@@ -0,0 +1,26 @@
...
+	/** strength: 8 to 12  */
+	Squad(2, 3),
+	/** strength: 3 to 4  */
+	Fireteam(1, 2),
+	/** strength: 2 to 3  */
+	FireAndManeuverTeam(0, 0)
```
Baret: I like "BuddyTeam" better :)

src/main/kotlin/de/gleex/pltcmd/model/elements/ElementImpl.kt
```diff
@@ -0,0 +1,19 @@
+package de.gleex.pltcmd.model.elements
+
+/**
+ * Implements an [Element] with a set of [Unit]s and subordinate [Element]s to whom orders are relayed.
+ */
+data class ElementImpl(val members: Set<UnitElement>, val subordinates: Set<Element>) : Element {
```
Baret: A default value for subordinates would be useful so you do not need to supply an emtpy list when creating UnitElements.

src/main/kotlin/de/gleex/pltcmd/model/elements/ElementImpl.kt
```diff
@@ -3,12 +3,12 @@ package de.gleex.pltcmd.model.elements
 /**
  * Implements an [Element] with a set of [Unit]s and subordinate [Element]s to whom orders are relayed.
  */
-data class ElementImpl(val members: Set<UnitElement>, val subordinates: Set<Element>) : Element {
+data class ElementImpl(val members: Set<Unit>, val subordinates: Set<Element>) : Element {
 
 	override fun executeOrder() {
```
Baret: This level of "executeOrder" is enough for now. Later this will work via `Command`s, so maybe it will be a whole different kind of mechanic. Let's not put too much work in Orders for now.

Loomie: That's why there is currently no parameter on this method ;)

Loomie:
>  Elements should know their parent, so you can traverse the hierarchy in both directions.
Sounds handy, so I will add it. This also changes the immutable implementation and will extend the interface to be able to edit elements.

> And another important thing that came to my mind is names and callsigns. This will be important gameplay wise, so we should add it early.
But in what way? I hate getters because they don't do anything. For data classes its ok, but a model class should do something useful. So I can think of a `reportCallsign()` method. But that is already possible with the generic `executeOrder()` method when providing a `ReportCallsignOrder` or something like that. I think that identifiers are only an internal thing which are relevant for addressing orders and for displaying the internal state to the user. ... After writing "relevant for addressing" it sound like I found a use case: A sender needs to know who can be adressed. So an `announceCallSign()` seems to be useful, because you can not send an order without an addressee. On the other hand there might be broadcasts without a specific addressee to which elements can reply...

src/main/kotlin/de/gleex/pltcmd/model/elements/ArmyElementHierarchy.kt
```diff
@@ -23,26 +26,41 @@ enum class ArmyElementHierarchy(val minCountOfConstituentUnits: Int, val maxCoun
...
 	
 	/** Creates a default Element with the average number of constituent units for every subordinate element. */
 	open fun createElement(superordinate: Element?): Element {
-		val element = ElementImpl(superordinate, setOf())
-		val consituentUnitType = values().getOrNull(ordinal + 1)
-		if (consituentUnitType != null) {
-			val averageCount = averageCountOfConsituentUnits()
+		val leader = GenericUnit(UnitType.Soldier)
+		val callSign = getCallSign(superordinate)
+		val element = Element(callSign, superordinate, setOf(leader))
```
Baret: cool kotlin stuff: `return Element(...).also { /* add subordinates */ }`. or `.apply{}`, I'm not sure what is the right choice here. But it's an optional change anyways ;)

src/main/kotlin/de/gleex/pltcmd/model/elements/Element.kt
```diff
@@ -1,10 +1,36 @@
...
+data class Element(val callSign: CallSign, val superordinate: Element?, val members: Set<Unit>) {
+	private val subordinates: MutableSet<Element> = mutableSetOf()
+
+	init {
+		// link from element to its subordinates
+		superordinate?.addSubordinate(this)
+	}
+
+	internal fun getSubordinates(): Set<Element> {
+		return subordinates.toSet()
```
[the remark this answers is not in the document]
Loomie: That's not possible, because the types are not compatible:
> Getter return type must be equal to the type of the property, i.e. 'MutableSet<Element>'

So I used a "[backing property](https://kotlinlang.org/docs/reference/properties.html#backing-properties)".

src/main/kotlin/de/gleex/pltcmd/model/elements/Element.kt
```diff
@@ -1,10 +1,36 @@
 package de.gleex.pltcmd.model.elements
 
+import java.lang.IllegalArgumentException
+
 /**
- * Part of a military hierarchy that follows orders.
+ * Part of a military hierarchy with a set of [Unit]s and subordinate [Element]s.
  */
-interface Element {
+data class Element(val callSign: CallSign, val superordinate: Element?, val members: Set<Unit>) {
```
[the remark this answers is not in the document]
Loomie: Instead of making it optional I have thought if `members` should be checked if they are empty. Because somebody must work in the element ;) But currently the units are not that important...

PR #17, opened 2020-01-04
Resolves #14 World Model

### LINKED ISSUE #14: Create the world
**Basic idea/big picture**
There needs to be a "world" containing all map tiles. There should be a way to access them in the form of _Sectors_ (see below).
Later the world will also hold all Elements (and other game entities) moving around. Probably this will be one part of a _Game_ class, that might hold additional state regarding the player.

**Details**
The world map consists of sectors. Each sector has 50 by 50 tiles. A Tile has a coordinate, for examlpe (577|488). The last two digits are the coordinates inside of sector (**5**[5x]|**4**[5x]). This means that two sectors make up a "big" coordinate (one sector from 500|400 to 549|400 and the next one from 550|400 to 599|400). But that could be used to display a world map, kind of "zooming out" of the default play view. I don't know if this will be needed though.
**For now it is enough to fill up the map with a default tile, representing a default TerrainHeight and TerrainType!** 

**Technical background**
Model:
I am not sure if the world map needs a map of sectors and each sector needs a map of tiles. Maybe it is enough to save _all_ the tiles in the world and just display 50 by 50 chunks of it. Or differently speaking: Just scroll the visible part of the map by 50 tiles. This could also make map generation (#13) easier.

UI:
The map should already be visible in the gameArea including some kind of border/decoration showing the coordinates. The "main coordinates" or "sector coordinates" should be highlighted. And there should be markings every 10 and 5 tiles, like shown [in the mockup](https://github.com/Baret/pltcmd/blob/master/mockups/mapview.png)

Loomie:


Baret:
So, finally I put together something that looks a bit like the final map and starts up :)
Although it still has lots of messy testing code, especially in the Sector-"generation", but that's to show all terrain heights and types. But please give it a review @Loomie so we can buldge the last bits and get it back to the master. Then we can start with proper map generation, which will sure bring some refactoring of parts of this branch with it.

src/main/kotlin/de/gleex/pltcmd/game/MapBlock.kt
```diff
@@ -0,0 +1,22 @@
...
+
+    companion object {
+        const val layersPerBlock = 3
+    }
+
+    private var unitTile: Tile = TileRepository.empty()
+    private val terrainTile: Tile = TileRepository.forType(terrain.type)
+    private val heightTile: Tile = TileRepository.forHeight(terrain.height)
+
+    override val layers: MutableList<Tile>
```
Loomie: A comment that describes the layers would be helpful. Maybe at class level to describe the order of the layers.

src/main/kotlin/de/gleex/pltcmd/options/GameOptions.kt
```diff
@@ -2,6 +2,12 @@ package de.gleex.pltcmd.options
 
 import de.gleex.pltcmd.model.radio.PercentageReducingAttenuation
 
+/**
+ * Options that change the behaviour of the game.
+ */
 object GameOptions {
     var attenuationModel = PercentageReducingAttenuation()
+
+    const val SECTORS_COUNT_H = 1
```
Loomie: not more?

src/main/kotlin/de/gleex/pltcmd/game/TileRepository.kt
```diff
@@ -0,0 +1,63 @@
...
+                withCharacter('\"').
+                build()
+        TerrainType.FOREST -> Tiles.newBuilder().
+                withForegroundColor(ColorRepository.forType(type)).
+                withBackgroundColor(TileColor.transparent()).
+                withCharacter(Symbols.SPADES).
+                buildCharacterTile()
+        TerrainType.HILL -> Tiles.newBuilder().
+                withForegroundColor(ColorRepository.forType(type)).
+                withBackgroundColor(TileColor.transparent()).
```
Loomie: A factory method would reduce the duplicated code, so here the code could become: `createTile(type, Symbols.X)`.

And then you realize that Grassland is not a character tile!?

src/main/kotlin/de/gleex/pltcmd/model/world/Sector.kt
```diff
@@ -0,0 +1,69 @@
...
+import de.gleex.pltcmd.model.terrain.TerrainHeight
+import de.gleex.pltcmd.model.terrain.TerrainType
+
+/**
+ * A sector has 50 by 50 [WorldTile]s (it is a square).
+ */
+data class Sector(val tiles: Set<WorldTile>) {
+	companion object {
+		/** edge length of a sector (in each directon of the map rectangle) */
+		const val TILE_COUNT = 50
```
Loomie: Could also be `GameOptions.SECTORS_COUNT_H`...

src/main/kotlin/de/gleex/pltcmd/model/world/Sector.kt
```diff
@@ -0,0 +1,69 @@
...
+			if (origin.eastingFromLeft % TILE_COUNT != 0 || origin.northingFromBottom % TILE_COUNT != 0) {
+				throw IllegalArgumentException("Origin of a Sector must be on a 50th of the map. Given: $origin")
+			}
+			// create all tiles for this sector
+			val sectorTiles = mutableSetOf<WorldTile>()
+			for (x in 0 until TILE_COUNT) {
+				for (y in 0 until TILE_COUNT) {
+					val coordinate = Coordinate(origin.eastingFromLeft + x, origin.northingFromBottom + y)
+					val height = TerrainHeight.values()[x % (TILE_COUNT / 5)]
+					var type = TerrainType.values()[(x / ( TILE_COUNT / 2) + y / (TILE_COUNT / 2))]
```
Loomie: A comment would be nice. Half of the sector gets a diagonal distribution of types and the other half fixed type 3?

src/main/kotlin/de/gleex/pltcmd/model/world/Sector.kt
```diff
@@ -0,0 +1,69 @@
...
+			// sectors must have full 50s as origin
+			if (origin.eastingFromLeft % TILE_COUNT != 0 || origin.northingFromBottom % TILE_COUNT != 0) {
+				throw IllegalArgumentException("Origin of a Sector must be on a 50th of the map. Given: $origin")
+			}
+			// create all tiles for this sector
+			val sectorTiles = mutableSetOf<WorldTile>()
+			for (x in 0 until TILE_COUNT) {
+				for (y in 0 until TILE_COUNT) {
+					val coordinate = Coordinate(origin.eastingFromLeft + x, origin.northingFromBottom + y)
+					val height = TerrainHeight.values()[x % (TILE_COUNT / 5)]
```
Loomie: Shouldn't  this be `% TerrainHeight.values().size` to prevent index out of bounds exceptions?

src/main/kotlin/de/gleex/pltcmd/options/GameOptions.kt
Baret: That was for testing purposes ;)
I had rendering issues and to make sure the visible part was where the sector was, I reduced it to 1.

src/main/kotlin/de/gleex/pltcmd/model/world/Sector.kt
Baret: I tried to create a sector showing all heights and all types. One terrainType on each quadrant of the sector and the height reducing from max to min from left to right. But at some point my modulo skills were done and I just didn't care anymore. It gets replaced by a nice map gerenator soon anyways, so this is just some showcase code.

src/main/kotlin/de/gleex/pltcmd/model/world/Sector.kt
Baret: Well... yes.

src/main/kotlin/de/gleex/pltcmd/model/world/Sector.kt
```diff
@@ -59,11 +60,11 @@ data class Sector(val tiles: Set<WorldTile>) {
 		return getSectorOrigin() == other.getSectorOrigin()
 	}
 
-	private fun WorldTile.getSectorOrigin(): Coordinate {
-		return Coordinate(
-			coordinate.eastingFromLeft - (coordinate.eastingFromLeft % TILE_COUNT),
-			coordinate.northingFromBottom - (coordinate.northingFromBottom % TILE_COUNT)
-		)
-	}
+	private fun WorldTile.getSectorOrigin() = coordinate.toSectorOrigin()
```
Loomie: An extension function that uses an extension function!? OK...

src/main/kotlin/de/gleex/pltcmd/model/world/MainCoordinate.kt
```diff
@@ -0,0 +1,13 @@
+package de.gleex.pltcmd.model.world
+
+/**
+ * Represents the "ain part" of a [Coordinate]. The main part consists of 100 by 100 tiles
+ * and is therefore a 100th of a coordinate (or "trnkating the last two digits").
```
Loomie: Typos: "ain part" and "trnkating". Also a dot at the end is missing.

src/main/kotlin/de/gleex/pltcmd/model/world/Coordinate.kt
```diff
@@ -7,4 +7,9 @@ package de.gleex.pltcmd.model.world
  *
  * It is like the numerical location of the Military Grid Reference System (see https://en.wikipedia.org/wiki/Military_Grid_Reference_System#Numerical_location).
  */
-data class Coordinate(val eastingFromLeft: Int, val northingFromBottom: Int)
\ No newline at end of file
+data class Coordinate(val eastingFromLeft: Int, val northingFromBottom: Int) {
+    /**
+     * Converts this coordinate to a [MainCoordinate]
+     */
+    fun toMaainCoordinate() = MainCoordinate(eastingFromLeft / 100, northingFromBottom / 100)
```
Loomie: -a

src/main/kotlin/de/gleex/pltcmd/model/world/Sector.kt
Baret: No. Or I do not understand what you mean. Those are two completely different constants.

src/main/kotlin/de/gleex/pltcmd/model/world/Sector.kt
Baret: Yes, they serve their own purpose (two different init checks). They are private for now, so it does not really matter I think.
Although Coordinate.toSectorOrigin might be handy later and be public or a member function of Coordinate.

Loomie:
Looks good now.

Baret:
I improved the formatting here and there. So now I think this branch can be merged. The small things missing (like the second MapDecorationRenderer) can be created in following issues. I really want this part to be in the master branch as it is a really good base for other things.

PR #27, opened 2020-01-25
Fixes #22: Flipped GameWorld in GameArea and GameArea only shows GameWorld

### LINKED ISSUE #22: Bug: GameWorld draws the WorldMap flipped
The `GameWorld` simply uses the values of the `Coordinates` to place the `MapBlocks`. But the map has 0,0 in the bottom left corner whereas the `GameArea` has its origin in the top left corner!

Loomie:
So the game starts with its block position in the top left corner at 0,0 but
the map grid starts in the bottom left corner at an arbitrary coordinate

### DIFF
--- a/src/main/kotlin/de/gleex/pltcmd/game/GameWorld.kt
+++ b/src/main/kotlin/de/gleex/pltcmd/game/GameWorld.kt
@@ -15,7 +15,7 @@ import org.hexworks.zircon.internal.game.impl.TopDownProjectionStrategy
  * The game world contains all [MapBlock]s representing the map. It has a visible part and can scroll from [Sector] to sector.
  * It is also capable of translating [Coordinate]s to [Position3D] and vice versa.
  */
-class GameWorld(worldMap: WorldMap):
+class GameWorld(private val worldMap: WorldMap):
         BaseGameArea<Tile, MapBlock>(
                 initialVisibleSize = Size3D.create(Sector.TILE_COUNT, Sector.TILE_COUNT, 1),
                 initialActualSize = Size3D.from2DSize(worldMap.size, 1),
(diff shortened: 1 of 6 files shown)


Loomie:
PR is mergable

src/main/kotlin/de/gleex/pltcmd/game/GameWorld.kt
```diff
@@ -18,7 +18,7 @@ import org.hexworks.zircon.internal.game.impl.TopDownProjectionStrategy
 class GameWorld(worldMap: WorldMap):
         BaseGameArea<Tile, MapBlock>(
                 initialVisibleSize = Size3D.create(Sector.TILE_COUNT, Sector.TILE_COUNT, 1),
-                initialActualSize = GameOptions.WORLD_SIZE,
```
Loomie: Should the constant be deleted?

Baret: I think we should keep `WORLD_SIZE`, because the generator needs to know how many sectors/tiles need to be created.

src/main/kotlin/de/gleex/pltcmd/game/GameWorld.kt
```diff
@@ -47,17 +50,39 @@ class GameWorld(worldMap: WorldMap):
...
-    private fun Position3D.toCoordinate() = Coordinate(x, y)
-
     /** Returns the [Coordinate] of the [Tile] that is visible in the top left corner. */
     fun visibleTopLeftCoordinate(): Coordinate {
         return visibleOffset.toCoordinate()
     }
 
-    fun scrollToCoordinate(coord: Coordinate) {
-        scrollTo(coord.toPosition())
+    fun scrollToCoordinate(bottomLeft: Coordinate) {
```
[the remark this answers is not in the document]
Loomie: > You could also build it more generic, [...] shifting the vieport to the sector containing the given Coordnate.
I think it is more generic, if we do not align the viewport to a world Sector. Else I would create another method: `scrollToSector()` ;)

PR #32, opened 2020-01-29
Resolves #13: Map Generators

### LINKED ISSUE #13: Map generator(s)
**Basic idea/big picture**
There should be a way to generate parts of the world map. The world map will be devided into _Sectors_. A sector is 50 by 50 tiles big. New sectors will be generated on the fly, so a map generator should at least be able to generate those 50*50 tiles.

**Details**
There should be different implementations of map generators to see what feels best. Or to choose a different generator for each sector. Some implementation suggestions are:
- Mountaintop(s): give the generator one or more positions of "mountain tops" (highest possible terrain height). The generator goes "downhill" from there in every direction. Every neighboring tile is even or lower, depending on a random value.
- BottomUp: The same as Mountaintop, but starting with the lowest terrain height.
- AllPlain: It gets a terrrain height and type and makes all tiles the same. A simlpe first implementation and maybe useful for debugging/testing
- AllRandom: Well, what it says on the box
- Smoothing: Running a different generator, for example the AllRandom, and then smooth the map a bit.

**Technical background**
As my first ideas are mostly based on height map generation, maybe it would be a good idea to split map generators into two parts. First a HeightMapGenerator is being run, setting only the TerrainHeight at all the tiles. After that a TerrainMapGenerator is called, putting the TerrainType on top of it. That way a more realistic vegetation might be achieved, like flowing rivers, mountains not lower than a specific TerrainHeight and so on.

At least #14 has to be done.

Loomie:


Baret:
I put down my idea of how map generation might work real quick (and dirty ;) ). It is just a rough idea and might be scratched completely. I am not yet sure if it is good to have several independent generators that only know which tile has been generated (or partly generated) yet but not more.

Baret:
I think we should set a reachable goal for this issue now, especially as I already wasted much time with stuff I wanted to play around with (...again ^.^).
I would consider this PR done when:
- there is more or less smooth terrain height everywhere, but not random and not plain
- there is mostly grasslands with a forest here and there and some (random) hill tiles

Further `IntermediateGenerators` can be implemented later (and hopefully easily). For example Lakes and rivers leading towards them. Or cities/urban areas.

Baret:
Wow, great perfomance boost thanks to the bug you found! 👍 
I didn't see your commits yet and worked on the same issue but wanted to tackle it with parallel streams. It didn't help that much. Then I pulled your 2 commits and was like O_O

I just now compared "your version" with "my version", meaning I added parallel streams back in and it made it a bit faster again. Although it seems buggy, as it does not create the full 2500.000 coordinates o.O

PR #40, opened 2020-02-25
resolves #36 use generation context

### LINKED ISSUE #36: Use the GenerationContext
The map generators are quite static right now. There is a `GenerationContext` but it is not being used by a generator. This should be fixed so that the MountainTopHeightMapper, for example, does not create the same amount of mountain tops every time.

Baret:
PR for issue #36


Baret:
btw I accidentially created this branch from #29 but I think as long as that one gets merged before this, it shouldn't be a problem.

Loomie:
All generators expect the RandomTerrainFiller use the context now. Only the values hilliness and vegetation are currently used.

Baret:
I'm reviewing...

src/main/kotlin/de/gleex/pltcmd/mapping/dijkstra/DijkstraMapOfCoordinates.kt
```diff
@@ -1,15 +1,116 @@
...
+import org.hexworks.cobalt.logging.api.LoggerFactory
+import kotlin.math.max
+import kotlin.math.min
+
 /**
  * not yet implemented!
+ * TODO: Fix documentation
  */
-class DijkstraMapOfCoordinates {
+class DijkstraMapOfCoordinates(private val map: Map<Coordinate, Int>) {
```
Baret: We will leave this class in that state. I created #50 to finish it.

src/main/kotlin/de/gleex/pltcmd/model/mapgenerators/GenerationContext.kt
```diff
@@ -2,54 +2,39 @@ package de.gleex.pltcmd.model.mapgenerators
...
+        val vegetation: Double,
+        val hilliness: Double,
         val water: Double,
-        val urban: Double,
-        /** open world to glue specific parts together */
-        val undefined: Double
+        val urban: Double
 ) {
+    init {
+        for (i in 1 until 3)
```
Baret: `vegetation` needs a triple check...? o_O

src/main/kotlin/de/gleex/pltcmd/model/mapgenerators/intermediate/HeightFiller.kt
```diff
@@ -60,8 +61,11 @@ private class PeekResult(
         private val toGenerate: Coordinate,
         private val peekedTile: Coordinate?,
         private val mutableWorld: MutableWorld,
-        private val rand: Random
+        private val rand: Random,
+        hilliness: Double
 ) {
+    val averageHeight = (TerrainHeight.MIN.value + (TerrainHeight.MAX.value - TerrainHeight.MIN.value) * hilliness)
```
Baret: Implement `TerrainHeight.times(Double)` and you can leave the `.value`. Maybe you also need `plus(Double)` though...
TerrainHeights are numbers internally so it should be possible to do calculations on them directly.

src/main/kotlin/de/gleex/pltcmd/model/mapgenerators/GenerationContext.kt
Loomie: Don't trust the `require()` :D

src/main/kotlin/de/gleex/pltcmd/model/mapgenerators/intermediate/HeightFiller.kt
Loomie: `TerreinHeight.times(Double)` cannot be implemented to calculate an average, because it is an enum with fixed values. So I cannot create `SIX.times(2.0)` to get TWELVE. It could return `TEN`. But that does not work for an accurate average. Thats because TerrainHeight is a set of integers and not a doubles.

Baret: For further calculations it could return a `Double` which can be turned back into a `TerrainHeight` by `forValue()`. But I see your point, it is not that important. I just hoped it would avoid future bugs when `TerrainHeight` would offer an API for calculations istead of trusting that you can rely on the inner intepretation of `value`. Now that I think about it, maybe it would have been good to make `value` private. Then we would have been forced to implement a bunch of `operator fun`s to do calculations and comparisons with `Double`s and `TerrainHeight`s (and maybe `Int`s).
Well, maybe when I'm in the right mood I will refactor that...

PR #45, opened 2020-03-15
View for map generation resolves #37

### LINKED ISSUE #37: View for Map generation
**Basic idea/big picture**
While the game generates the map it would be interesting to see how it forms and evolves. The process takes some seconds so it might be possible to "spectate" it.

**Details**
Currently the game starts in the `GameView` which is only visible after map generation has finished. There could be a view displaying an abstract version of the `MutableWorld` while it is being generated, so you have some kind of loading screen. I imagine it being full window size, maybe with a border titled "generating world...". It juts has a `GameAreaComponent` which maps chunks of the `MutableWorld` to displayed tiles, which means when the window is 100 tiles wide and the world being generated has 1000 tiles in width, it would display the average type and height (or empty) of 10 tiles at each position.

**Technical background**
The new view would need to observe the `MutableWorld` and map generation would need to run asynchonously. So far I have 2 ideas of how this could be implemented, but better ones might exist ;)
- The view gets the `MutableWorld` as some kind of [observable property](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.properties/-delegates/observable.html) (though the native kotlin way might not be useful here). Maybe it would already be enough to let `MutableWorld` extend `BaseGameArea` and let Zircon do the rest.
- Every `IntermediateGenerator` sends an Event (i.e. over the Cobalt `EventBus`) after a step that might justify a redraw of the map. But this approach would need every future `IntermediateGenerator` to respect that kind of contract, which is not nice as it might be forgotten.

As we just finished #13 this might be started any time.

Loomie:
- Added title screen.
- Showing progress of map generation
- Display minimap of world while generating


src/main/kotlin/de/gleex/pltcmd/model/radio/RadioSignal.kt
```diff
@@ -40,12 +40,12 @@ open class RadioSignal(private val power: Double) {
...
      */
-    fun along(terrain: List<Terrain>): SignalStrength {
+    fun along(terrain: List<Terrain?>): SignalStrength {
         if(terrain.size <= 1) {
             return power.toSignalStrength()
         }
-        val startHeight = terrain.first().height
-        val targetHeight = terrain.last().height
+        val startHeight = terrain.first()?.height ?: TerrainHeight.TEN
+        val targetHeight = terrain.last()?.height ?: TerrainHeight.ONE
```
Baret: First of all there is `TerrainHeight.MIN/MAX`. But why should a radio signal travel along nullable terrain?! It travels through the world which is... well, present. Not void.

src/main/kotlin/de/gleex/pltcmd/model/radio/RadioSignal.kt
```diff
@@ -55,6 +55,8 @@ open class RadioSignal(private val power: Double) {
                 // currentHeight (y) = mx + b
                 val currentHeight = floor(slope * (index + 1) + startHeight.toDouble())
                 signal = when {
+                    // signal travels through the void (not yet generated terrain)
```
Baret: "not yet generated terrain" is not a case that will ever happen in the game. That is what I mean with the mixed architecture when you misuse `GameWorld` and co to display a mutable world. We need to add strange workarounds like that only because of a part that happens _before_ the actual game.

src/main/kotlin/de/gleex/pltcmd/PreviewGenerationListener.kt
```diff
@@ -0,0 +1,78 @@
+package de.gleex.pltcmd
```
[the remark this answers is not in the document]
Loomie: The documentation says: "shows". Meaning that the UI is involved. So it is a bridge between the model and the UI.

src/main/kotlin/de/gleex/pltcmd/PreviewGenerationListener.kt
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+}
+
+/** Holds all real world data that is mapped onto a single preview tile */
+data class AverageTerrain(val mappedTiles: MutableMap<Coordinate, Pair<TerrainHeight?, TerrainType?>> = mutableMapOf()) {
+
+    @Suppress("MemberVisibilityCanBePrivate")
+    fun getAverageHeight(): TerrainHeight? {
+        val average = mappedTiles.values.stream()
```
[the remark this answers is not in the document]
Loomie: Because I wanted `mapToInt()` to get an `IntStream`. I didn't found that in Kotlin. Nor was there an `average()` for lists.

src/main/kotlin/de/gleex/pltcmd/PreviewGenerationListener.kt
```diff
@@ -0,0 +1,78 @@
...
+    private fun mapToPreview(coordinate: Coordinate): Coordinate {
+        val easting = (coordinate.eastingFromLeft - previewWorld.topLeftOffset.x) / tilesPerPreviewWidth
+        val northing = (coordinate.northingFromBottom - previewWorld.topLeftOffset.y) / tilesPerPreviewHeight
+        return Coordinate(easting, northing)
+    }
+
+}
+
+/** Holds all real world data that is mapped onto a single preview tile */
+data class AverageTerrain(val mappedTiles: MutableMap<Coordinate, Pair<TerrainHeight?, TerrainType?>> = mutableMapOf()) {
```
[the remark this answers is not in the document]
Loomie: Seemed so special. But now it looks like a normal class. You're right!

src/main/kotlin/de/gleex/pltcmd/PreviewGenerationListener.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    @Suppress("MemberVisibilityCanBePrivate")
+    fun getAverageHeight(): TerrainHeight? {
+        val average = mappedTiles.values.stream()
+                .filter { it.first != null }
+                .mapToInt { it.first!!.value }
+                .average()
```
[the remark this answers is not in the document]
Loomie: I moved a bit of code to `TerrainHeight`. But here is still a conversion necessary from Pair to `TerrainHeight`

src/main/kotlin/de/gleex/pltcmd/PreviewGenerationListener.kt
```diff
@@ -0,0 +1,78 @@
...
+        if (!average.isPresent()) {
+            return null
+        }
+        val averageHeightValue = average.asDouble.toInt()
+        return TerrainHeight.ofValue(averageHeightValue)
+    }
+
+    @Suppress("MemberVisibilityCanBePrivate")
```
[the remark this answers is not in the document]
Loomie: Yeah, but then there are warnings 🙄
I wanted to state: This is a public API, because I think it is useful - fuck you IDEA!

src/main/kotlin/de/gleex/pltcmd/model/radio/RadioSignal.kt
Loomie: I removed the optional Terrain from RadioSignal. But the MapBlock may still be empty.

src/main/kotlin/de/gleex/pltcmd/main.kt
```diff
@@ -25,4 +34,28 @@ fun main() {
...
+
+private fun generateMap(screen: Screen, tileGrid: TileGrid): GameWorld {
+    val origin = Coordinate(0, 0)
+    val mapGenerator = WorldMapGenerator(GameOptions.DEBUG_MAP_SEED)
+
+    val previewWorld = createPreview(origin)
+    val generatingView = GeneratingView(previewWorld, tileGrid)
+    screen.dock(generatingView)
+
+    val progressListener = ProgressListener(mapGenerator.sizeInTiles, 100.0, generatingView.progressProperty) // 100.0 is the default of ProgressBar
```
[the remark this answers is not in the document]
Loomie: I added the comment because else the value for the Property does not make much sense. The Property could be mapped, but with such an indirection I can also use a listener to notify about the changed value.

src/main/kotlin/de/gleex/pltcmd/PreviewGenerationListener.kt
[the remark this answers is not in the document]
Baret: Yup, some classes might belong to different packages. But the two listerners are implementations of `MapGenerationListener`. In my opinion that makes them more belong to the `mapgenerators` package. Now that I say it, maybe the package should be renamed to `mapgeneration` as it contains more than just the generators themself. And probably it also doesn't need to be under `model` but one level higher, next to `game` (which contains completely different stuff. The only interconnection between those two packages should be that the `MutableWorld` creates a `GameWorld`).

src/main/kotlin/de/gleex/pltcmd/PreviewGenerationListener.kt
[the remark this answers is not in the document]
Baret: I think there are special extension functions for Int and Double lists. And corresponding to that you can of course also create an extension function for List<TerrainHeight>.

src/main/kotlin/de/gleex/pltcmd/PreviewGenerationListener.kt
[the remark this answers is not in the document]
Baret: Valid point! I also thought about setting up different modules once and if it was already too late to do that. If you have a good idea for a possible structure go ahead and create an issue and write down your proposal there.

src/main/kotlin/de/gleex/pltcmd/model/mapgenerators/ui/PreviewGenerationListener.kt
```diff
@@ -0,0 +1,36 @@
...
+
+import de.gleex.pltcmd.game.GameWorld
+import de.gleex.pltcmd.model.mapgenerators.MapGenerationListener
+import de.gleex.pltcmd.model.terrain.AverageTerrain
+import de.gleex.pltcmd.model.terrain.TerrainHeight
+import de.gleex.pltcmd.model.terrain.TerrainType
+import de.gleex.pltcmd.model.world.Coordinate
+import org.hexworks.zircon.api.data.Position3D
+
+/** Shows the current state of the world in a single sector sized [GameWorld] while the map is beeing generated. */
```
Baret: This KDoc is outdated.

src/main/kotlin/de/gleex/pltcmd/model/mapgenerators/ui/PreviewGenerationListener.kt
```diff
@@ -0,0 +1,36 @@
...
+class PreviewGenerationListener(generatedWorldWidth: Int, generatedWorldHeight: Int, private val previewWorld: IncompleteMapGameArea) : MapGenerationListener {
+    private val displaySize = previewWorld.actualSize.to2DSize()
+    private val tilesPerPreviewWidth = generatedWorldWidth / displaySize.width
+    private val tilesPerPreviewHeight = generatedWorldHeight / displaySize.height
+
+    private val averageTerrainMap: MutableMap<Position3D, AverageTerrain> = mutableMapOf()
+
+    override fun terrainGenerated(coordinate: Coordinate, terrainHeight: TerrainHeight?, terrainType: TerrainType?) {
+        val previewCoordinate = mapToPreview(coordinate)
```
Baret: This is not a coordinate but a position. And the method name is not very descriptive imo. Why not make it an extension function that says "this coordinate is being converted to a `Position3D`? Makes it much more readable.

src/main/kotlin/de/gleex/pltcmd/model/mapgenerators/ui/PreviewGenerationListener.kt
```diff
@@ -0,0 +1,36 @@
...
+
+        val terrainData = averageTerrainMap.getOrPut(previewCoordinate, { AverageTerrain() })
+        terrainData.mappedTiles[coordinate] = Pair(terrainHeight, terrainType)
```
Baret: I think it is not nessecary to make `mappedTiles` public. There is no need to fiddle around with `Pair` outside of `AverageTerrain` (it can keep its inner data sctructures for itself).
```suggestion
        terrainData.put(coordinate, terrainHeight, terrainType)
```
(I thought about a nicer syntax for this call but everything else that came to my mind wasn't more readable or easier to use)

src/main/kotlin/de/gleex/pltcmd/model/mapgenerators/ui/IncompleteMapGameArea.kt
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    fun setBlock(x: Int, y: Int, terrainHeight: TerrainHeight?, terrainType: TerrainType?) {
+        val position = Position3D.create(x, y, 0)
+        setBlock(position, terrainHeight, terrainType)
+    }
+
+    fun setBlock(position: Position3D, terrainHeight: TerrainHeight?, terrainType: TerrainType?) {
+        val block = IncompleteMapBlock(terrainHeight, terrainType)
+        setBlockAt(position, block)
```
Baret: Maybe it would be more performant if you updated the content of the block instead of replacing it with another one. Showing different tiles in its lifetime is exactly the intention of a Block in Zircon (i.e. player moves through the level in a roguelike).
As it works for now you may just keep it. But if you want, give it a shot and try shorter cycles in the preview-update-executor (or remove it alltogether) and see if you still have the issues that were the reason to invent it in the first place.

src/main/kotlin/de/gleex/pltcmd/model/terrain/AverageTerrain.kt
```diff
@@ -0,0 +1,39 @@
...
+    fun createTerrain(): Terrain? {
+        val averageHeight = getAverageHeight()
+        val dominatingType = getDominatingType()
+        if (averageHeight == null && dominatingType == null) {
+            return null
+        }
+        val type = dominatingType ?: TerrainType.WATER_DEEP
+        val height = averageHeight ?: TerrainHeight.MIN
+        return Terrain.of(type, height)
+    }
```
Baret: This method is obsolete and can be removed.

src/main/kotlin/de/gleex/pltcmd/model/terrain/AverageTerrain.kt
```diff
@@ -0,0 +1,39 @@
+package de.gleex.pltcmd.model.terrain
+
+import de.gleex.pltcmd.model.world.Coordinate
+import java.util.*
+import java.util.stream.Collectors
+
+/** Holds all real world data that is mapped onto a single preview tile */
+data class AverageTerrain(val mappedTiles: MutableMap<Coordinate, Pair<TerrainHeight?, TerrainType?>> = mutableMapOf()) {
+
+    @Suppress("MemberVisibilityCanBePrivate")
```
Baret: I think now we can remove those annotations as the only two methods of this class obviously are the only API it offers. And they are both used so your IDE should not complain ;)

src/main/kotlin/de/gleex/pltcmd/ui/GeneratingView.kt
```diff
@@ -0,0 +1,64 @@
...
+
+    override fun onDock() {
+        val header = createHeader()
+        val mainPart = createMainPart()
+
+        screen.addComponents(header, mainPart, progressBar)
+    }
+
+    private fun createHeader(): Label {
+        return Components.label()
```
Baret:
```suggestion
    private fun createHeader(): Header {
        return Components.header()
```
1. It does what it says it would ;)
2. It is much more readable. When I started it up the first couple of times I didn't even notice there was text above the map.

src/main/kotlin/de/gleex/pltcmd/ui/GeneratingView.kt
```diff
@@ -0,0 +1,64 @@
...
+                .withSize(incompleteWorld.visibleSize.xLength, incompleteWorld.visibleSize.yLength - usedLines)
+                .withPosition(0, headerHeight)
+                .build()
+
+        val map = GameComponents.newGameComponentBuilder<Tile, IncompleteMapBlock>()
+                .withGameArea(incompleteWorld)
+                .withSize(mainPart.size)
+                .withAlignmentWithin(mainPart, ComponentAlignment.CENTER)
+                .build()
+        mainPart.addComponent(map)
```
Baret: Is it actually nessecary to wrap the `GameComponent` into a `Panel`? It should have have the needed properties to directly set size and position on it I think.

src/main/kotlin/de/gleex/pltcmd/model/mapgenerators/ui/PreviewGenerationListener.kt
Loomie: You're right, that the inner field should not be accessible.

src/main/kotlin/de/gleex/pltcmd/ui/GeneratingView.kt
Loomie: Yes, it is unnecessary. I think it was required in the first version of Zircon that we used. But GameComponent is now more integrated.

Loomie: Ah, the probem in `main` is that is adds Decorations. Also there the visibleArea is used for scrolling. A short try to remove the panel there resulted in graphic artifacts (the map was only visible in the decoration)

Loomie:
Now there is a "Click to continue" button that replaces the progress bar.

src/main/kotlin/de/gleex/pltcmd/model/mapgenerators/data/MutableWorld.kt
```diff
@@ -109,24 +115,27 @@ class MutableWorld(val bottomLeftCoordinate: Coordinate = Coordinate(0, 0),
...
-        terrainMap[coordinate] = Pair(terrainHeight, terrainType)
+        terrainMap.getOrPut(coordinate) { TerrainData() }
+                .update(terrainHeight, terrainType)
         fireChange(coordinate, terrainHeight, terrainType)
     }
 
     private fun fireChange(coordinate: Coordinate, terrainHeight: TerrainHeight?, terrainType: TerrainType?) {
-        eventExecutor.execute { listeners.forEach { it.terrainGenerated(coordinate, terrainHeight, terrainType) } }
+        // remember current listeners for async call
+        val listenersToNotify = HashSet(listeners)
```
Baret:
```suggestion
        val listenersToNotify = listeners.toSet()
```

src/main/kotlin/de/gleex/pltcmd/model/mapgenerators/ui/IncompleteMapGameArea.kt
```diff
@@ -6,33 +6,35 @@ import org.hexworks.zircon.api.data.Position3D
...
+ * A [GameArea] that shows a map with holes. It uses [IncompleteMapBlock] for that.
+ * The full map is visible and can be altered.
+ */
 class IncompleteMapGameArea(size: Size) :
         BaseGameArea<Tile, IncompleteMapBlock>(
                 initialVisibleSize = Size3D.from2DSize(size, 1),
                 initialActualSize = Size3D.from2DSize(size, 1),
-                initialContents = mapOf(),
-                projectionStrategy = TopDownProjectionStrategy()) {
+                initialContents = initialContents(size.width, size.height)) {
```
Baret:
```suggestion
                initialContents = size.
                                    fetchPositions().
                                    map { it.to3DPosition(0) }.
                                    associateWith { IncompleteMapBlock() }) {
```

src/main/kotlin/de/gleex/pltcmd/model/mapgenerators/ui/IncompleteMapGameArea.kt
```diff
@@ -6,33 +6,35 @@ import org.hexworks.zircon.api.data.Position3D
...
 
-    init {
-        // fill area with voidness
-        for (y in 0 until size.height) {
-            for (x in 0 until size.width) {
-                setBlock(x, y, null, null)
+    companion object {
+        private fun initialContents(width: Int, height: Int): Map<Position3D, IncompleteMapBlock> {
```
Baret: This stacked loop feels too complicated for kotlin. There is almost always a niftier solution ;)

src/main/kotlin/de/gleex/pltcmd/ui/GeneratingView.kt
```diff
@@ -5,60 +5,90 @@ import de.gleex.pltcmd.model.mapgenerators.ui.IncompleteMapGameArea
...
 import org.hexworks.zircon.api.data.Size
 import org.hexworks.zircon.api.data.Tile
+import org.hexworks.zircon.api.game.GameComponent
 import org.hexworks.zircon.api.grid.TileGrid
 import org.hexworks.zircon.api.view.base.BaseView
 
-/** Displays the progress of world generation. A miniature of the world is shown together with a progress bar. */
+/**
+ * An incomplete map is shown together with a progress bar. Both must be altered externally. After calling
+ * [showFinished] a button will shown that triggers a callback when used.
```
Baret:
```suggestion
 * [showFinished] a button will become visible that triggers a callback when used.
 * @see onConfirmation to set the callback for the button.
```

src/main/kotlin/de/gleex/pltcmd/ui/GeneratingView.kt
```diff
@@ -5,60 +5,90 @@ import de.gleex.pltcmd.model.mapgenerators.ui.IncompleteMapGameArea
...
+ **/
 class GeneratingView(tileGrid: TileGrid) : BaseView(theme = UiOptions.THEME, tileGrid = tileGrid) {
 
+    private val footer = createFooter()
     private val progressBar = createProgressBar()
```
Baret:
```suggestion
    private val progressBar = createProgressBar()
    private val footer = createFooter().apply{ addComponent(progressBar) }
```

src/main/kotlin/de/gleex/pltcmd/ui/GeneratingView.kt
```diff
@@ -5,60 +5,90 @@ import de.gleex.pltcmd.model.mapgenerators.ui.IncompleteMapGameArea
...
+    private val usedLines = progressBar.height + header.height
+    private var confirmCallback: () -> Unit = {}
+
     val incompleteWorld = IncompleteMapGameArea(Size.create(screen.width, screen.height - usedLines))
 
-    override fun onDock() {
-        val header = createHeader()
+    init {
         val mainPart = createMainPart()
+        footer.addComponent(progressBar)
```
Baret:
```suggestion
```

src/main/kotlin/de/gleex/pltcmd/model/mapgenerators/ui/IncompleteMapGameArea.kt
Loomie: nice. Didn't knew about `associateWith` 👍

src/main/kotlin/de/gleex/pltcmd/model/mapgenerators/WorldMapGenerator.kt
```diff
@@ -14,7 +14,7 @@ import kotlin.random.Random
  * The WorldMapGenerator is the only class you need outside this package. It generates the world for the game given a seed.
  */
 class WorldMapGenerator(
-        private val seed: Long,
+        private val seed: Long = GameOptions.DEBUG_MAP_SEED,
```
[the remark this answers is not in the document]
Loomie: The same counts for the size. Generating a world without providing in which dimensions should not be possible.

PR #55, opened 2020-04-13
Fixes #48: TypeFiller fills terrain without requirering TerrainHeight.MAX

### LINKED ISSUE #48: Bug: TypeFiller does not work if no terrain with TerrainHeight.MAX exists
**Description**
The TypeFiller does nothing if it cannot start with a mountain top. Therefore all terrain will get filled with random TerrainType. This occurs if GenerationContext.hilliness is low (as of #36).

**Location**
TypeFiller.generateArea() line 24ff. The `processedTiles` must not be empty for the `workFrontier()` call.

**Reproduction**
Set `hilliniess` to 0.0 in the GenerationContext.

Loomie:


Loomie:
Was not based on master :(

PR #58, opened 2020-04-24
Resolves #19: Basics with Amethyst

### LINKED ISSUE #19: Basic system for GameEntities (aka "use amethyst")
**Basic idea/big picture**
Now that there is a world and it can be displayed, it is time to introduce [Amethyst](https://github.com/hexworks/amethyst).

**Details**
As this issue should only create the very basic system for future game mechanics, it is enough to have a _game_ that contains an _engine_ and puts one unit somewhere on the map and displays it.

**Technical background**
Basics on how to implements `Entity`s, their `Attribute`s and `System`s to handle them can be [found here](https://hexworks.org/posts/tutorials/2019/02/13/how-to-make-a-roguelike-the-player.html).

#14 should be done and in the master before starting.

Baret:
Solution for #19


Loomie:
Elements with a position are now entities. They can move to a destination by a behaviour. The destination is set by a command and executed in a facet. All entities are updated by the `GameEngine` that is triggered every second in the main.

model/world/src/main/kotlin/de/gleex/pltcmd/model/world/WorldMap.kt
```diff
@@ -31,8 +31,11 @@ data class WorldMap(val sectors: Set<Sector>) {
      * Returns all neighbors of the given coordinate that are inside this world.
      */
     fun neighborsOf(coordinate: Coordinate): List<Coordinate> {
-        // TODO: Needs check if all neighbors are inside the world (probably c&p code from MutableWorld...)
-        return coordinate.neighbors()
+        return coordinate.neighbors().filter { contains(it) }
```
Baret, line 34: This is ok, but highly inefficient because in the worst case you check every single coordinate.
We can improve this later by letting `WorldMap` and `Sector` inherit from `CoordinateRectangle` and putting a bounds check like in `MutableWorld` there (where you simply check x/y of origing and topLeftCoordinate).

Issue #31, opened 2020-01-27
Fix map decorations


Baret:
**Basic idea/big picture**
Press cursor keys to move to the next sector. Maybe a scrolling via mouse should also be possible, for example by clicking the map border

**Details**
The easy part is: Pressing on of the cursor keys moves the visible map one sector east/west/north/south, if present.
Tricky might be to build a good system to make the map scrollable with the mouse. As "border scrolling" (by just moving the mouse to the map border) like in RTS games would not be a viable option, it should be triggered via mouse click. Maybe the inner decoration could show a clickable arrow under the mouse cursor when hovering over it.

**Technical background**
This issue might be fundamental for User input.

**Related issues/Prerequisites**
To scroll to other sectors, there have to be multiple sectors. So the map needs to be filled with them, ideally because #13 is fixed and they are generated. But for test purposes it would be enought to fill up with the "debug sectors".

Baret:
This has been implemented in #13 but has the bug that the map decorations do not get updated when scrolling. So this issue should be used to fix it. Maybe there is more to improve though...

PR #66, opened 2020-05-07
Fixes 31 map scrolling and decorations

Baret:
PR for #31


Baret:
The actual fix for this is the simple line `gameWorld.visibleOffsetValue.onChange { asInternalComponent().render() }`
I think the way Zircon works we have to trigger a re-rendering of the whole map component (or rather it's surrounding panel) because decorations only get rendered, when the component renders. I did not find a way to let a decoration "redraw itself".
I was thinking about data-binding, so that the decoration would get a property bound to the visibleOffset and all displayed coordinateStrings would update accoding to it, but that would have been quite complicated without knowing if it worked.

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/extensions/EntityExtensions.kt
```diff
@@ -9,5 +9,5 @@ import kotlin.reflect.KClass
  *
  * Be sure the attribute exists when using.
  */
-internal fun <T : Attribute> AnyGameEntity.getAttribute(attribute: KClass<T>) =
+fun <T : Attribute> AnyGameEntity.getAttribute(attribute: KClass<T>) =
```
Loomie, line 12: it was intentional internal - all attributes should only be accessed by the enity (incl. its behaviours and facets). Also no new code uses this. So I will revert this change.

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/GameView.kt
```diff
@@ -43,7 +43,9 @@ class GameView(private val gameWorld: GameWorld, tileGrid: TileGrid, val element
                 withSize(UiOptions.MAP_VIEW_WDTH, UiOptions.MAP_VIEW_HEIGHT).
                 withAlignmentWithin(screen, ComponentAlignment.TOP_RIGHT).
                 withDecorations(MapGridDecorationRenderer(), MapCoordinateDecorationRenderer(gameWorld)).
-                build()
+                build().apply {
+                    gameWorld.visibleOffsetValue.onChange { asInternalComponent().render() }
```
Loomie, line 47: A comment would be helpful.

PR #70, opened 2020-05-20
Issue #65 rework elements model

Baret:
I started playing around with an idea of mine regarding the model for units and elements (#65). The basics are complete, now I add some sugar to it.


model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/reworked/CommandingElement.kt
```diff
@@ -0,0 +1,115 @@
...
+        get() {
+            return if(superordinate.isPresent) {
+                superordinate.get().callSignFor(this)
+            } else {
+                ownCallsign
+            }
+        }
+
+    private fun callSignFor(subordinate: Element): CallSign {
+        val index = subordinates.indexOf(subordinate)
```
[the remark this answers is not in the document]
Loomie: Although it is unordered, the iteration order is normally fix. So it might be OK for a single JVM run ;)

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/reworked/Element.kt
```diff
@@ -0,0 +1,81 @@
+package de.gleex.pltcmd.model.elements.reworked
+
+import de.gleex.pltcmd.model.elements.reworked.units.Unit
+import org.hexworks.cobalt.core.platform.factory.UUIDFactory
+import org.hexworks.cobalt.datatypes.Maybe
+
+/**
+ * Elements make up a  military hierarchy. Each element groups a set of [Unit]s.
```
Loomie, line 8: The elements do not make up the hierarchy. That is what ElementSize is for. That is the formal aspect. But elements are the executing part. They make up the armed forces or just "the military".

Instead of "groups a set of Units" I would say "consist of a set of Units" or similar as the people make up the element. As I see it the Element is just a name (kind/size) for the group of soldiers.

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/reworked/Element.kt
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * Simple elements are commanded by a superordinate (see [CommandingElement]).
+ * They are not part of the command net and do not have an own callsign.
+ *
+ * Every element has a [kind] and [size] defining what kind of units they are composed of
+ * and on what level of organisation they reside (the bigger, the further up in the hierarchy).
+ */
+open class Element(
+        /**
+         * @see ElementKind
```
Loomie, line 18: That doesn't help much as the type is already given. You could add this type of comment to all parameters and variables...

Loomie: Oh, if the documentation of `ElementKind` is for `Element.kind` than put it here ;)

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/reworked/units/Unit.kt
```diff
@@ -0,0 +1,51 @@
...
+ * A _unit_ is one independent piece of "military equipment" (i.e. a soldier or a weapon system or a vehicle)
+ * on the field. It has a unique ID and the following fields:
+ *
+ * [name] describing this unit, for example "Rifleman".
+ *
+ * [personnel] is the default number of soldiers that operate this unit at full strength.
+ *
+ * [personnelMinimum] is the number of soldiers needed to keep this unit operational.
+ *
+ * [kind] see [UnitBlueprint.kind]
```
Loomie, line 16: There is no documentation on that field 👎 I think the documentation should be directly here where it belongs!

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/reworked/units/Unit.kt
```diff
@@ -0,0 +1,51 @@
...
+ * be present once. This class represents the immutable part of a unit. Mutable attributes should be added by the game
+ * engine to track its state.
+ */
+class Unit(val blueprint: UnitBlueprint) {
+    val id = UUIDFactory.randomUUID()
+    /**
+     * @see [Unit]
+     */
+    val name: String
+        get() = blueprint::class.objectInstance?.javaClass?.simpleName?: "Unit"
```
Loomie, line 28:
```suggestion
        get() = blueprint::class.simpleName ?: "Unit"
```

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/reworked/units/Unit.kt
```diff
@@ -0,0 +1,51 @@
...
+ * [personnelMinimum] is the number of soldiers needed to keep this unit operational.
+ *
+ * [kind] see [UnitBlueprint.kind]
+ *
+ * Units are backed by a [UnitBlueprint] for memory optimization. All immutable base values of a unit only need to
+ * be present once. This class represents the immutable part of a unit. Mutable attributes should be added by the game
+ * engine to track its state.
+ */
+class Unit(val blueprint: UnitBlueprint) {
+    val id = UUIDFactory.randomUUID()
```
Loomie, line 23: Can be used for `equals()/hashCode()`. But as it is immutable and each object gets a unique id each object is unique anyway. So there is no need to implement equals :)

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/reworked/ElementSize.kt
```diff
@@ -0,0 +1,36 @@
...
+     * rather unlikely to. They are usually too cumbersome to maneuver effectively in fight and should only
+     * be used to move big numbers of troops around the map.
+     */
+    Company,
+    /**
+     * A battalion is a huge group of units in context of this game. It is unlikely to ever exist except as
+     * an organizational bucket to keep lots of elements in reserve. Most probably this size is just out of
+     * scope of the player's paycheck ;)
+     */
+    Battalion
```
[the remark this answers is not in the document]
Baret: Honestly I cant remember. Do you mean something like `CommandigElement.allUnits` or `.soldierCount`? (Not sure about the exact names but CE has a field for "how many soldiers")

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/reworked/units/Unit.kt
Baret: I wanted to add equals using id later ;)

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/reworked/ElementKind.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import de.gleex.pltcmd.model.elements.reworked.units.Unit
+import de.gleex.pltcmd.model.elements.reworked.units.UnitKind
+
+/**
+ * The kind of an [Element] determines its icon representation and which kinds of units are allowed.
+ *
+ * @see [UnitKind]
+ */
+enum class ElementKind(val allows: Set<UnitKind>) {
```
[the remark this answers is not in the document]
Baret: I tried making it private but the error message in `Element.addUnit()` uses it for a verbose error message and maybe it might be handy in some other place later to directly access the set. So I just renamed it.

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/reworked/units/UnitKind.kt
```diff
@@ -0,0 +1,14 @@
...
+ * [Unit]s are of different kind. Foot soldiers belong to infantry, trucks are unarmored and main battle tanks are
+ * heavily armored, for example.
+ */
+enum class UnitKind {
+    Infantry,
+    Unarmored,
+    ArmoredLight,
+    ArmoredHeavy,
+    AerialLight,
+    AerialHeavy
```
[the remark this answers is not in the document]
Baret: I thought about naval units but came to the conclusion that they are "too special". Armed forces may combine land, air and sea units but often naval units are a different branch of the army.
But more important the current map generation does not offer maps with enough water areas that it would make sense. Maybe we can add it in later but for now I think that lakes (not yet present) are just natural borders. And rivers may either be crossed by most units when shallow or we put bridge building units into the logistics corps.

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/reworked/CommandingElement.kt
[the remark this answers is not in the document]
Baret: I thought about keeping a mapping of all subordinates and their callsign suffixes first. But it seemed overengineered and as I discovered that you can use an index in a set it was good enough for me.
But now I found a bug: When adding a fourth subordinate, it gets the suffix "-4". When you remove one of the first 3 elements the callsing changes to "-3". I think this is not good. As long as an element is attached to another one it should keep its callsign. I will fix that.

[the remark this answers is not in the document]
Baret: I Implemented a `SubCallSignProvider` now. Keeping track if the callsign isn't as easy as I thought, but for now it works. There are special edge cases but I think for now we could keep it like it is.
When you execute the elementsExample and remove the "when they are back on their own" part, you see that "Red Wolfs" becomes "Alpha-4" and stays "Alpha-4" when one of the elements 1 to 3 are removed. But when Alpha gets the new superordinate "Wild Hogs", they become "Wild Hogs-1-4" instead of 1-3.
As I don't know how often this might happen and if the player actually sees those "sub callsigns" I accept that slight misbehaviour.

Baret:
I fixed or awnsered all your comments. Thanks for the great feedback! You can review it again now, but I still have some more things I would like to finish in this PR.

- Use the corps. It will basically be another restriction when adding elements to elements (corps needs to be the same).
- Make an enum containing all units so that we can easily programatically create an exhaustive list of ingame content (i.e. a little "unit database" screen).
- Create unittests. I would basically create tests from the example cases I already wrote

game/application/src/main/kotlin/de/gleex/pltcmd/game/application/examples/elementsReworked/elementsExamples.kt
```diff
@@ -0,0 +1,136 @@
...
+    println("First it looks like this:")
+    printCommandElement(wolfs)
+    spacer()
+    println("But after adding it to ${alpha.callSign}...")
+    val elementToRemoveFromAlpha = alpha.subordinates.random() as CommandingElement
+    alpha.addElement(wolfs)
+    printCommandElement(wolfs)
+    println("Directly adding it to bravo should not be possible")
+    tryAndCatch {
+        bravo.addElement(wolfs)
```
Loomie, line 55: When printing the whole company the wolfs are listed beneath bravo! They have their own name so they seem to not have a superordinate, but they seem to be a subordinate of bravo!

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/reworked/CommandingElement.kt
```diff
@@ -0,0 +1,120 @@
...
+    private fun callSignFor(subordinate: Element): CallSign =
+            callsignProvider.callSignFor(subordinate)
+
+    /*{
+        val index = subordinates.indexOf(subordinate)
+        require(index >= 0) {
+            "Element $subordinate is not a subordinate of $this"
+        }
+        return callSign + "-${index + 1}"
+    }*/
```
Loomie, line 82: This commented code an be removed now?!

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/reworked/Element.kt
```diff
@@ -0,0 +1,109 @@
...
+ * They are not part of the command net and do not have an own callsign.
+ *
+ * Every element has a [kind] and [rung] defining what kind of units they are composed of
+ * and on what level of organisation they reside.
+ */
+open class Element(
+        val kind: ElementKind,
+        val rung: Rung,
+        units: Set<Unit>,
+        superordinate: CommandingElement? = null
```
Loomie, line 20: Looks odd that a class gets it subclass as parameter. Maybe use an interface instead!? Or just use `Element` here to indicate another element. There should be a check that superordinate is not set to this object!

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/reworked/SubCallsignProvider.kt
```diff
@@ -0,0 +1,65 @@
...
+                } else {
+                    (callsignMap.values.max() ?: 0) + 1
+                }
+    }
+
+    private fun deletionHappened(change: ObservableValueChanged<PersistentSet<Element>>) =
+            change.newValue.size < change.oldValue.size
+
+    private fun additionHappened(change: ObservableValueChanged<PersistentSet<Element>>) =
+            change.newValue.size > change.oldValue.size
```
Loomie, line 57: A replacement of an element will not be detected! Although it is an addition and deletion. Maybe the size check plus `old != new`.

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/reworked/CommandingElement.kt
Baret: Oops...

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/reworked/Element.kt
Baret: I found it strange too, but it makes sense. A `CommandingElement` _is_ an element and a superordinate needs to be a commanding element. I do not see how an interface would make this code cleaner, as it would just be a marker interface on a subclass of `Element`.

But the check is definately a good idea!

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/reworked/ElementSize.kt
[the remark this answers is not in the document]
Loomie: I meant [these kind of comments](https://github.com/Baret/pltcmd/blob/master/model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/ArmyElementHierarchy.kt#L17)
> strength: 80 to 250

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/reworked/SubCallsignProvider.kt
Baret: Yes, there is a logical case missing here. But I think the `PersistentSet` does not allow replacement (i.e. put) of elements. It just has add and remove, but I will check.

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/reworked/ElementSize.kt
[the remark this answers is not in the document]
Baret: done

Baret:
After fiddling around with the callsign and its inheritance by the superordinate we (after some discussion) decided to keep it simple for now. I will change the model to the following:
- An element gets a property containing its superordinate. This way it is observable and when putting it from one CommandingElement to another it can properly be updated
- The callsing of a CommandingElement stays fixed. No more inheritance as it is too complicated to keep track of and probably won't be used later.

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/reworked/SubCallsignProvider.kt
Baret: ...finally done: 38147cf
I solved it with a `require` but I also tried a validator in the property:
```kotlin
private var _superordinate: Property<Maybe<CommandingElement>> =
            initialSuperOrdinate
                    .toProperty {
                        if (it.isPresent) {
                            it.get() != this
                        } else {
                            true
                        }
                    }
```
But this just doesn't update the value in case the validation returns `false`. I found this behaviour too intransparent and rather wanted to throw an exception ("dead programs tell no lies").

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/reworked/Element.kt
Baret: ...finally done: 83147cf 
I solved it with a `require` but I also tried a validator in the property:
```kotlin
private var _superordinate: Property<Maybe<CommandingElement>> =
            initialSuperOrdinate
                    .toProperty {
                        if (it.isPresent) {
                            it.get() != this
                        } else {
                            true
                        }
                    }
```
But this just doesn't update the value in case the validation returns `false`. I found this behaviour too intransparent and rather wanted to throw an exception ("dead programs tell no lies").

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/reworked/SubCallsignProvider.kt
Baret: The code for an observebale set (aka `SetProperty`) now went over to `CommandingElement` as `_subordinates` is now being watched for changes. And there I tried to implement the "official" or I hope at least "correct" way of implementing it. And it looks aweful...! I didn't know how to go about the `ChangeType` and found an example in `ListBinding` ([see here](https://github.com/Hexworks/cobalt/blob/master/cobalt.core/src/commonMain/kotlin/org/hexworks/cobalt/databinding/internal/binding/ListBinding.kt#L34)).
This is probably totally overkill as we control what operations may happen on the set (`addElement()` and `removeElement()`). But I wanted to learn how it is intended :)

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/blueprint/ElementBlueprintStart.kt
```diff
@@ -0,0 +1,35 @@
...
+     * or to create a commanding element by chaining [ElementBlueprint.commanding].
+     */
+    infix fun consistingOf(unit: Units) = consistingOf(listOf(unit))
+}
+
+/**
+ * To start declaring an [ElementBlueprint] you first need an [ElementBlueprintStart]. It sets
+ * the basic values for the resulting element.
+ */
+internal fun a(corps: Corps, kind: ElementKind, rung: Rung) =
```
Loomie, line 34: Hm, looks like a type alias whould do the same thing. I wonder if this extension function is more or less helpful 🤔 The extension function can/must be imported explicitly whereas a type alias is global for a package (I think). But types _should_ be named upper case and functions lower case so this might be more readable!?

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/units/Units.kt
```diff
@@ -0,0 +1,223 @@
...
+ * Use [new] to create new instances of a unit. Or use arithmetic operators to create sets of [Unit]s.
+ * For example "2 * Rifleman + Grenadier".
+ *
+ * Each unit should come with a documentation explaining the intention of how the unit works and
+ * explaining abbreviations.
+ */
+enum class Units(
+        val kind: UnitKind = Infantry,
+        val personnel: Int = 1,
+        val personnelMinimum: Int = 1
```
Loomie, line 22: There should be a validation in the `init` to check that `personal` is at least `personnelMinimum`.

model/elements/src/test/kotlin/de/gleex/pltcmd/model/elements/units/UnitsCalculationsTest.kt
```diff
@@ -0,0 +1,4 @@
+package de.gleex.pltcmd.model.elements.units
+
+class UnitsCalculationsTest {
```
Loomie, line 3: Pretty empty if not to say useless.

util/namegeneration/src/main/kotlin/de/gleex/pltcmd/util/namegeneration/Namegenerator.kt
```diff
@@ -0,0 +1,11 @@
+package de.gleex.pltcmd.util.namegeneration
+
+/**
+ * A name generator returns randomly generated names. They should be unique as long as possible.
+ */
+interface Namegenerator {
+    /**
+     * Generates a new random name.
```
Loomie, line 8: instead of  "random" it should provide a "unique" name.

model/elements/src/test/kotlin/de/gleex/pltcmd/model/elements/ReasonableDefaultElementsTest.kt
```diff
@@ -0,0 +1,81 @@
...
+                row(Elements.engineerSquad, infantrySquadSize),
+                row(Elements.engineerPlatoon, infantryPlatoonSize)
+        ) { elementBlueprint, (min, max) ->
+            "The default ${Elements.nameOf(elementBlueprint)}" should {
+                "have between $min and $max soldiers" {
+                    val newElement = elementBlueprint.new()
+                    val soldierCount = if (newElement is CommandingElement) {
+                        newElement.totalSoldiers
+                    } else {
+                        newElement.units.sumBy { it.personnel }
```
Loomie, line 74: Why isn't there a `totalSoldiers` on `Element`?

model/elements/src/test/kotlin/de/gleex/pltcmd/model/elements/blueprint/CommandingElementBlueprintTest.kt
```diff
@@ -0,0 +1,79 @@
...
+                        context("Each squad ($i)") {
+                            squad as CommandingElementBlueprint
+                            assertUnits(squad)
+                            should("should only contain ${ElementBlueprint::class.simpleName}s") {
+                                squad.subordinates.forAll {
+                                    it should beInstanceOf(ElementBlueprint::class)
+                                }
+                                squad.subordinates.forNone {
+                                    it should beInstanceOf<Element>()
+                                    it should beInstanceOf<CommandingElementBlueprint>()
```
Loomie, line 46: Interesting that the generic is enough for testing instanceof. In the above tests (line 28f) the class was provided.

model/elements/src/test/kotlin/de/gleex/pltcmd/model/elements/blueprint/CommandingElementBlueprintTest.kt
```diff
@@ -0,0 +1,79 @@
...
+                                }
+                            }
+
+                            squad.subordinates
+                                    .forEachIndexed { i, fireteam ->
+                                        context("Fireteam ($i)") {
+                                            should("should only have unit blueprints, no units") {
```
Loomie, line 53: This should also be extraced to `assertUnits` - if CommandingElementBlueprint inherits from ElementBlueprint the one function can be used for both ;)

model/elements/src/test/kotlin/de/gleex/pltcmd/model/elements/blueprint/ElementsDslTest.kt
```diff
@@ -0,0 +1,59 @@
...
+                        CommandingElementBlueprint(Corps.Fighting, ElementKind.Infantry, Rung.Squad, listOf(Units.Rifleman), listOf(ElementBlueprint(Corps.Fighting, ElementKind.Infantry, Rung.Fireteam, listOf(Units.Rifleman)))))
+        ) { createdBlueprint: CommandingElementBlueprint, expectedBlueprint: CommandingElementBlueprint ->
+            createdBlueprint shouldBe expectedBlueprint
+        }
+    }
+
+    val blueprint = a(Corps.CombatSupport, ElementKind.Armored, Rung.Battalion) consistingOf 5 * Units.MainBattleTank
+    "An ${ElementBlueprint::class.simpleName} should create an ${Element::class.simpleName}" {
+        blueprint should beInstanceOf(ElementBlueprint::class)
+        blueprint.new() should beInstanceOf(Element::class)
```
Loomie, line 47: Not so different from the first "should result in the correct element blueprint".

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/CommandingElement.kt
```diff
@@ -0,0 +1,164 @@
...
+                            val addedElements = mutableSetOf<Element>()
+                            when(type) {
+                                is ScalarChange         -> {
+                                    if(oldSet.size > newSet.size) {
+                                        removedElements.addAll(oldSet - newSet)
+                                    } else {
+                                        addedElements.addAll(newSet - oldSet)
+                                    }
+                                }
+                                is ListChange -> {
```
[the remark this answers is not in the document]
Baret: Jup, exactly my thoughts. But I wanted to see how this changeType thingy was intended to work. An I was rather shocked :D

Just kept it in for you to see, now I am going to simplify it.

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/Blueprint.kt
```diff
@@ -0,0 +1,11 @@
+package de.gleex.pltcmd.model.elements
+
+/**
+ * A blueprint used to create new instances of [T].
+ */
+interface Blueprint<out T> {
```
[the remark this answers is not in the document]
Baret: I recognised that it was very generic, too. But I didn't want to create a "util.reallyGenericUtilStuff" module just for a single interface. So I will keep it here for now.

And for the name: Blueprint is much more precise. Factory sounds like something that produces many different objects, a blueprint is for exactly one thing.
I just thought about renaming it and that I would need to rename the blueprint classes, too. But `UnitFactory` just woudln't fit imo.

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/CallSignProvider.kt
```diff
@@ -0,0 +1,39 @@
...
+
+    // TODO: Pick name generator based on corps, kind and rung
+    private val generator: Namegenerator = when(corps) {
+        Corps.Fighting  -> AlphabetPicker() // Cool names
+        Corps.Logistics -> AlphabetPicker() // Animal names
+        // something like that...
+        else            -> AlphabetPicker()
+    }
+
+    private var callSign: CallSign = CallSign(generator.generate())
```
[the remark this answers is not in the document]
Baret: Honestly I didn't know where to put the "pick the correct name generator" code. Becase after #84 there will be different generators that know nothing about `ElementKind` or `Corps`. So we need to decide which one to use here.
Yes, it might be a static field, but would always need to get the corps, rung and kind as parameters. Well, just like the `CallSignProvider` does now.
At least there might be a "repick name" mechanism later which would justify keeping the generator. On the other hand one could just re-instantiate one then.
I will see how I can improve this...

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/CommandingElement.kt
```diff
@@ -0,0 +1,164 @@
...
+                                        is ListRemoveAllWhen<*> -> removedElements.addAll(oldSet.filter { oldItem -> (type.predicate as (Element) -> Boolean)(oldItem) })
+                                    }
+                                }
+                            }
+                            addedElements.forEach {
+                                it.superordinate = Maybe.of(this@CommandingElement)
+                            }
+                            removedElements
+                                    .filter { it.superordinate.isPresent }
+                                    .filter { it.superordinate.get() == this@CommandingElement }
```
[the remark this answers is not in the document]
Baret: I think there was a case when this made sense but I can not exactly remember what it was...
Maybe I first had a data binding in both direcitons: The superordinate reset `subordinate.superordinate` which in turn called `superordinate.removeElement(this)` and I had to break this cycle. Or something like that :P

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/Element.kt
```diff
@@ -1,38 +1,116 @@
...
+    /**
+     * If this element is currently being commanded this [Maybe] contains the superordinate.
+     *
+     * When this elements gets a new superordinate it automatically removes itself from the
+     * former superordinate (if it was present).
+     */
+    var superordinate: Maybe<CommandingElement>
+        get() = _superordinate.value
+        set(value) {
+            if(value.isPresent) {
```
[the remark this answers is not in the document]
Baret: Because I thought the validator would silently ignore the invalid value and so I decided to throw an exception instead.
I believe it returns some kind of validation result though, so maybe I could react to that.

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/Elements.kt
```diff
@@ -0,0 +1,146 @@
...
+                    .declaredMemberProperties
+                    .filter { it.visibility == KVisibility.PUBLIC }
+                    .filter {it.returnType.isSubtypeOf(ElementBlueprint::class.starProjectedType) }
+                    .associate { it.name to it.getter.call(this) as ElementBlueprint }
+
+    /**
+     * All available blueprints. For specific variants see [allCommandingElements] and [allElements]
+     */
+    fun all(): Map<String, Blueprint<Element>> =
+            allElements() + allCommandingElements()
```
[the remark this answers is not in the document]
Baret: Umm... there sure was a reason I couldn't subclass it. But I already realized that the two blueprint classes need a common  superclass (ideally `sealed`) so I can get rid of those `Blueprint<Element>` instance checks and unify those calculation operator funs.

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/Elements.kt
```diff
@@ -0,0 +1,146 @@
...
+
+    /**
+     * Tries to find the name of the given blueprint by searching [Elements.all].
+     */
+    fun nameOf(blueprint: Blueprint<Element>): String? {
+        return all().entries.firstOrNull { it.value == blueprint }?.key
```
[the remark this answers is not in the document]
Baret: Yes, another thought I already had. But I wanted to describe the elements with a nicely to read "DSL" instead of calling those (different) constructors over and over again (quite like in `Units`, actually... :\ ).

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/Elements.kt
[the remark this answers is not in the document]
Baret: Ah, I think it was because one is a `Blueprint<Element>` and the subclass would be a `Blueprint<CommandingElement>`, which needs `new()` to return a different class. Now that i write it down, I am not sure if the abstract superclass works. But I will try :)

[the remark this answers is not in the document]
Baret: ...generics, fuck yeah! :)

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/Elements.kt
```diff
@@ -0,0 +1,146 @@
...
+    val weaponsSquad =
+            a(Fighting, Infantry, Squad) consistingOf squadLead commanding 2 * weaponsTeam
+
+    val antiTankSquad =
+            a(Fighting, Infantry, Squad) consistingOf squadLead commanding 2 * antiTankTeam
+
+    val antiAirSquad =
+            a(Fighting, Infantry, Squad) consistingOf squadLead commanding 2 * antiAirTeam
+
+    val motorizedInfantryTeam =
```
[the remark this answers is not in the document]
Baret: The order was rather "by corps" and then "by ElementKind", which I messed up :D

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/Elements.kt
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * This is just the commanding element of a platoon without subordinates. "An empty platoon".
+     */
+    val fightingInfantryPlatoonCommand =
+            a(Fighting, Infantry, Platoon) consistingOf platoonLead commanding emptyList()
+
+    val riflePlatoon =
+            fightingInfantryPlatoonCommand commanding 3 * rifleSquad
+
+    val heavyInfantryPlatoonHMG =
```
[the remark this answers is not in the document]
Baret: I was thinking of the icon those elements would get and imagined a "heavy infantry" icon. So I introduced that "group" on the platoon level and added the name of the heavy weapons system contained.
But I changed it so that the rung stays at the last position.

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/Elements.kt
```diff
@@ -0,0 +1,146 @@
...
+            a(Fighting, Infantry, Squad) consistingOf squadLead commanding 2 * antiAirTeam
+
+    val motorizedInfantryTeam =
+            a(Fighting, MotorizedInfantry, Fireteam) consistingOf 3 * Rifleman + Grenadier
+
+    val motorizedTransportSquad =
+            a(Fighting, MotorizedInfantry, Squad) consistingOf 1 * TransportTruck + Officer
+
+    val motorizedInfantrySquad =
+            // TODO: Add something like rifleSquad.withKind(MotirizedInfantry) to reuse existing blueprints
```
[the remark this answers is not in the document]
Baret: We need to keep a balance between reusablility and readability here. I think it is ok to have some `a(Logistics, MotorizedInfantry, Fireteam)`. In the end I wouldn't even know how to call the composition of corps and kind.
Now that blueprints can be mutated to different kind or corps it is much more tidy already :)

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/blueprint/ElementBlueprint.kt
```diff
@@ -0,0 +1,41 @@
...
+        /**
+         * Visible for testing.
+         */
+        internal val units: List<Units>
+): Blueprint<Element> {
+    override fun new() = Element(
+                                    corps,
+                                    kind,
+                                    rung,
+                                    units.map(Units::new).toSet()
```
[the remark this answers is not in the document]
Baret: It is a set to make sure it is a collection of unique units (unique by their IDs).

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/blueprint/ElementBlueprintStart.kt
Baret: A `typealias` is an alias for a _type_ but here we call the _constructor_ of a type. Or in other words: A typealias can not have parameters. So there is that.

Baret: Oh, now I got it :D
`internal typealias A = ElementBlueprintStart` resulting in `A(Fighting, Infantry, Fireteam)` which looks strange imho. So I keep it like that.

util/namegeneration/src/main/kotlin/de/gleex/pltcmd/util/namegeneration/Namegenerator.kt
Baret: Yes, of course. But I wanted to leave the details of this interface and its implementations to #84.

model/elements/src/test/kotlin/de/gleex/pltcmd/model/elements/CommandingElementTest.kt
```diff
@@ -0,0 +1,192 @@
...
+                    .filter { it == commandingElement.corps }
+            "allow adding elements of corps $validCorps when it is of corps ${commandingElement.corps} itself" {
+                validCorps.forAll {
+                    val subElement = buildSubElement(corps = it)
+                    checkSubordination(commandingElement, subElement)
+                }
+            }
+
+            "have been checked against all ${Corps.values().size} corps" {
+                invalidCorps + validCorps shouldHaveSize Corps.values().size
```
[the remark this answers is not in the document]
Baret: Sometimes I am getting paranoid when writing tests... :D

model/elements/src/test/kotlin/de/gleex/pltcmd/model/elements/CommandingElementTest.kt
```diff
@@ -0,0 +1,192 @@
...
+                // create new CE, check all subs
+                val ce = newCE()
+                val sub1 = buildSubElement()
+                val sub2 = buildSubElement()
+                val sub3 = buildSubElement()
+                val allSubs = setOf(sub1, sub2, sub3)
+
+                ce.subordinates should beEmpty()
+                allSubs.forAll {
+                    it.superordinate should de.gleex.pltcmd.util.tests.beEmpty()
```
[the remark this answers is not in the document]
Baret: It has to because exactly 2 lines above I use another `beEmpty()` (coming from `io.kotest.matchers.collections`). Trust the IDE when organizing imports ;)

model/elements/src/test/kotlin/de/gleex/pltcmd/model/elements/ElementTest.kt
```diff
@@ -0,0 +1,59 @@
...
+import io.kotest.inspectors.forExactly
+import io.kotest.matchers.collections.shouldContain
+import io.kotest.matchers.collections.shouldHaveSize
+import io.kotest.matchers.collections.shouldNotContain
+import io.kotest.matchers.shouldBe
+
+class ElementTest : WordSpec({
+
+    val rifleman = Units.Rifleman.new()
+    val simpleElement = Element(Corps.Reconnaissance, ElementKind.MotorizedInfantry, Rung.Fireteam, setOf(Units.Officer.new(), rifleman, rifleman))
```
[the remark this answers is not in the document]
Baret: No, I do not want a Collection of `Unit`s in an element. I want a set.
But yes, maybe I am just testing `setOf` here...

model/elements/src/test/kotlin/de/gleex/pltcmd/model/elements/ElementTest.kt
```diff
@@ -0,0 +1,59 @@
...
+    "A simple element $simpleElement" should {
+        "contain 2 units" {
+            simpleElement.units shouldHaveSize 2
+            simpleElement.units.forExactly(1) { it.blueprint shouldBe Units.Officer }
+            simpleElement.units.forExactly(1) { it.blueprint shouldBe Units.Rifleman }
```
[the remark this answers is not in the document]
Baret: Instead of an equality check on a set (`shouldBe setOf(...)`) I think one should rather use `shouldContainExactly`.

model/elements/src/test/kotlin/de/gleex/pltcmd/model/elements/ElementTest.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    // Test element creation with invalid unit kinds
+    ElementKind.values()
+            .forEach { elementKind ->
+                val invalidUnitKinds = UnitKind
+                        .values()
+                        .filter {
+                            elementKind.allowedUnitKinds.contains(it).not()
+                        }
+                invalidUnitKinds.forEach { invalidUnitKind ->
```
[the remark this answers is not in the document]
Baret: Yes, I was thinking about it, too. But it just is more descriptive. You immediately see what the filtered list contains and do not need to read and understand the `filter` statement.

model/elements/src/test/kotlin/de/gleex/pltcmd/model/elements/ReasonableDefaultElementsTest.kt
```diff
@@ -0,0 +1,81 @@
...
+                        .forEach { entry ->
+                            log.info("\t${entry.key}:")
+                            val blueprint = entry.value
+                            log.info("\t\t$blueprint")
+                            if (blueprint is CommandingElementBlueprint) {
+                                blueprint.subordinates.forEach { sub ->
+                                    log.info("\t\t${0x251C.toChar()} ${Elements.nameOf(sub)}")
+                                }
+                            }
+                            blueprint shouldBe instanceOf(CommandingElementBlueprint::class).or(instanceOf(ElementBlueprint::class))
```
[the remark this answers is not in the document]
Baret: I added a new test for the actual types further down. And renamed this one.

model/elements/src/test/kotlin/de/gleex/pltcmd/model/elements/ReasonableDefaultElementsTest.kt
```diff
@@ -0,0 +1,81 @@
...
+                row(Elements.heavyInfantryPlatoonAT, infantryPlatoonSize),
+                row(Elements.heavyInfantryPlatoonAA, infantryPlatoonSize),
+                row(Elements.motorizedInfantryPlatoon, infantryPlatoonSize),
+
+                row(Elements.engineerTeam, infantryFireteamSize),
+                row(Elements.engineerSquad, infantrySquadSize),
+                row(Elements.engineerPlatoon, infantryPlatoonSize)
+        ) { elementBlueprint, (min, max) ->
+            "The default ${Elements.nameOf(elementBlueprint)}" should {
+                "have between $min and $max soldiers" {
```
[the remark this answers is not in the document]
Baret: The test is called "reasonable defaults" because it is meant to be a reminder for anyone who adds new default elements or changes existing ones. It should just pop up when you change the rifleTeam to suddenly have 20 riflemen (by accident or bug) which is pretty sure incorrect.
The actual number of units in an element depends on many factors and so putting it in `inits`s and `required`s would be so much hard coded stuff it would become ugly. For example the scout plane is perfectly fine to be alone but maybe on the rung of Squad and even be a commanding element (to be represented on the radio net). But an infantry platoon should be "around 3 to 4 squads" depending on the actual composition.

model/elements/src/test/kotlin/de/gleex/pltcmd/model/elements/ReasonableDefaultElementsTest.kt
Baret: Now that you say it... ;P

model/elements/src/test/kotlin/de/gleex/pltcmd/model/elements/blueprint/CommandingElementBlueprintTest.kt
Baret: Oh yes, I missed chaning that. I tried both and though this one would be better.

model/elements/src/test/kotlin/de/gleex/pltcmd/model/elements/blueprint/ElementsDslTest.kt
Baret: No, the first test says "the DSL should create the correct _blueprint_. This test say a so created blueprint should create the correct _element_.

Baret:
I hope I got all your comments. As you can see I put my changes in maaaany many tiny commits :) Almost one for every comment.

Go ahead.

model/elements/src/test/kotlin/de/gleex/pltcmd/model/elements/ElementTest.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    val officer = Units.Officer.new()
+    val rifleman = Units.Rifleman.new()
+    val simpleElement = Element(Corps.Reconnaissance, ElementKind.MotorizedInfantry, Rung.Fireteam, setOf(officer, rifleman))
+    "A simple element $simpleElement" should {
+        "contain 2 units" {
+            simpleElement.units shouldHaveSize 2
+            simpleElement.units.forExactly(1) { it.blueprint shouldBe Units.Officer }
+            simpleElement.units.forExactly(1) { it.blueprint shouldBe Units.Rifleman }
+            simpleElement.units shouldContainExactlyInAnyOrder setOf(officer, rifleman)
```
Loomie, line 24: the three checks above are now redundant: it contains exactly two elements which is the `officer` of type Units.Officer and `rifleman` of type Units.Rifleman`

model/elements/src/test/kotlin/de/gleex/pltcmd/model/elements/ReasonableDefaultElementsTest.kt
```diff
@@ -0,0 +1,156 @@
...
+                row(Elements.transportHelicopterSquad, 2 to 2),
+                row(Elements.transportHelicopterPlatoon, 6 to 9),
+                row(Elements.transportTruckTeam, 1 to 2),
+                row(Elements.transportTruckSquad, 2 to 4),
+                row(Elements.transportTruckPlatoon, 14 to 18)
+        ) { elementBlueprint, (min, max) ->
+            "Unit count: The default ${Elements.nameOf(elementBlueprint)}" should {
+                "have between $min and $max units" {
+                    val newElement = elementBlueprint.new()
+                    val soldierCount = if (newElement is CommandingElement) {
```
Loomie, line 100:
```suggestion
                    val unitCount = if (newElement is CommandingElement) {
```

model/elements/src/test/kotlin/de/gleex/pltcmd/model/elements/ReasonableDefaultElementsTest.kt
```diff
@@ -0,0 +1,156 @@
...
+                        newElement.units.size
+                    }
+                    soldierCount.shouldBeBetween(min, max)
+                }
+            }
+        }
+
+        "All default elements" should {
+            "be sorted" {
+                Elements.all().keys.toList()
```
Loomie, line 112: Why are you calling `all()` again and again if you already have the variable `allElements`? ;)

model/elements/src/test/kotlin/de/gleex/pltcmd/model/elements/ReasonableDefaultElementsTest.kt
```diff
@@ -0,0 +1,156 @@
...
+                        .shouldBeSorted()
+                Elements.allElements().keys.toList()
+                        .shouldBeSorted()
+                Elements.allCommandingElements().keys.toList()
+                        .shouldBeSorted()
+            }
+
+            "have a name" {
+                Elements.all()
+                        .map { it.key }
```
Loomie, line 122: is this map different to `.keys`?

model/elements/src/test/kotlin/de/gleex/pltcmd/model/elements/ReasonableDefaultElementsTest.kt
```diff
@@ -0,0 +1,156 @@
...
+                val publicFields = Elements::class
+                        .declaredMemberProperties
+                        .filter { it.visibility == KVisibility.PUBLIC }
+
+                val publicBlueprints: List<AbstractElementBlueprint<*>> = publicFields
+                        // we explicitly check for the subtypes of AbstractElementBlueprint
+                        // because it does not help when one of the fields is of the abstract type
+                        .filter {
+                            it.returnType.isSubtypeOf(ElementBlueprint::class.starProjectedType)
+                                    || it.returnType.isSubtypeOf(CommandingElementBlueprint::class.starProjectedType)
```
Loomie, line 146: Almost like the implementation. So the test code compares with the same implementation code - not a great benefit.
A hard coded list here would be better on the expected side.

model/elements/src/test/kotlin/de/gleex/pltcmd/model/elements/blueprint/ElementsDslTest.kt
Loomie: Yeah and that is a test for the blueprint implementation but not so much for the DSL which is tested here.

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/blueprint/AbstractElementBlueprint.kt
```diff
@@ -0,0 +1,154 @@
...
+            units.map(Units::new)
+                    .toSet()
+    )
+
+    override fun times(i: Int) =
+            List(i) { ElementBlueprint(corps, kind, rung, units) }
+
+    override fun withKind(newKind: ElementKind) = ElementBlueprint(corps, newKind, rung, units)
+
+    override fun withCorps(newCorps: Corps) = ElementBlueprint(newCorps, kind, rung, units)
```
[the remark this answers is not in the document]
Baret: Yup I also missed the `copy()` here. But the abstract class may not be a data class. I dint try it on the subclasses though. Then the `with*` funs could simply be abstract.

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/CommandingElement.kt
```diff
@@ -0,0 +1,146 @@
...
+        units: Set<Unit>,
+        subordinates: Set<Element> = emptySet()
+) : Element(corps, kind, rung, units) {
+
+    private val _subordinates: SetProperty<Element> =
+            mutableSetOf<Element>()
+                    .toProperty()
+                    .apply {
+                        onChange { (oldSet, newSet, _, _) ->
+                            if (oldSet.size > newSet.size) {
```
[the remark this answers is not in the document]
Baret: You are right. I removed the condition. But I did not put it in seperate methods as this code is now down to 4 lines and that is just because I left the comments in there.

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/Elements.kt
```diff
@@ -0,0 +1,182 @@
...
+    // ######################
+    // Recon corps
+    // ######################
+
+    // ======================
+    //      Aerial (AIR)
+    // ======================
+
+    val reconPlane =
+            a(Reconnaissance, Aerial, Fireteam) consistingOf ScoutPlane commanding emptyList()
```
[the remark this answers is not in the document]
Baret: First off the rung has nothing to do with the possibility to be represented on the command net. A sniper team for example might be only two guys but you definetely want to control them directly, so it is just a fireteam but with a callsign.

Maybe I can think about the name `CommandingElement` once more but the intention is as follows. There are elements that are too small to effectively execute orders/missions on their own (classic example is the infantry fireteam or even buddy team). These are "basic elements" that are not controllable by the commander and thus are not represented on the command net (aka they have no call sign). They are simply controlled locally by their superordinate which is not relevant from the commaner's perspective.
Then there are elements "big" or "valuable" enough to be a tool in the commander's toolbox. They can be used to fulfil an intent and to complete missions. So in my opinion only those need to have a call sign. They may have subordinates or they may be on their own but they are "a marker on the map" that the commander pushes around.

And on a side note: Generating a call sign for every single element in the faction (even though it is never being used) raises the requirements for name generators drastically as they simply need to have a larger pool of names to pick from.

I will however add another method to create an empty `CommandingElement` so you do not need to pass `emptyList()`.

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/blueprint/AbstractElementBlueprint.kt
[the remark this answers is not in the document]
Baret: Ah, right:
> Data class primary constructor must have only property (val / var) parameters

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/blueprint/AbstractElementBlueprint.kt
```diff
@@ -0,0 +1,154 @@
+package de.gleex.pltcmd.model.elements.blueprint
+
+import de.gleex.pltcmd.model.elements.*
+import de.gleex.pltcmd.model.elements.units.Units
+import de.gleex.pltcmd.model.elements.units.new
+
+/**
+ * Base class for the element blueprints.
+ */
+sealed class AbstractElementBlueprint<T : Element>(
```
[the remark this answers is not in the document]
Baret: There must not be any "own implementations". This is exactly the sense of `sealed`. It says "there are exactly these two implementations".
We need a blueprint for `Element`s and a blueprint for `CommandingElement`. But to nicely handle "any blueprint" (like in calculation operator funs) I needed a base class.

It makes things [like this possible](https://github.com/Baret/pltcmd/pull/70/commits/a95c18eda408c0662e963f6894c10eb1273dd062#diff-32821e0de1399568b57987ae9e30089aR44) ;)

Yes, the file got rather large but I think it is ok as it exactly fits the use case.

model/elements/src/test/kotlin/de/gleex/pltcmd/model/elements/ReasonableDefaultElementsTest.kt
Baret: The reason I came to this (and also the other `shouldBe instanceOf` test further up) is the history of development... :)

When creating the operator funs I had one return an `AbstractElementBlueprint` so one of the default elements didn't have the type `ElementBlueprint` or `CommandingElementBlueprint`. I had a test to count `Elements.all()` but it just didn't pick up fields of other types than the two concrete blueprint classes. That's why I created a test that finds all fields of those types and compares it to `Elements.all()`.
The intention is: All public `val`s in `Elements` should either be an `ElementBlueprint` or a `CommandingElementBlueprint` o you can be sure every field creates either an `Element` or a `CommandingElement`. When we later add further defaults it should pop up as "design fail" when we end up adding a public field of the wrong type (or, as already encountered, if the "DSL" results in the wrong return type ^^ ).

Baret:
Strangely I can not reply to your last review comment o_O

Well, maybe I missed an explicit `ElementBlueprintTest` ^^ The test you mentioned is kind of a mix. The blueprint is created using the DSL so it tests both. And regarding that I left it there, because else I would simply copy and paste the test and just create the blueprint using the constructor. This way I got both parts tested.

Again, everything answered or solved I hope. Merge it already! ;D

model/elements/src/test/kotlin/de/gleex/pltcmd/model/elements/ElementTest.kt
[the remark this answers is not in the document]
Loomie: But the variable for this lambda is named `invalidUnitKind`

PR #80, opened 2020-06-21
Basic Combat Resolves #79

### LINKED ISSUE #79: Basic Combat
**Basic idea/big picture**
Elements fight each other to eleminate the enemy. The elements of different factions are attacking each other on sight/reach.

**Details**
Each element (or sub elements) needs a fighting power and health condition. In the first implementation the attack can just reduce the health by its amount. For example each element gets 100 health and and an attack of 20. After receiving five attacks the element is eleminated. Also each element should have a fixed range for attacking other elements, maybe one to begin with.

This would already bring in basic tactics to pull elements back before they get killed or provide a superiority of units to ensure victory.

**Technical background**
The fighting power can be fixed or already depend on the health. The health will most probably be an attribute of an ElementEntity. But maybe this can already be encapsulated in an ElementState that is extended later.

The detection of nearby enemies needs the map for the surrounding coordinates. Then there must be a way to get other elements in the target area. For those elements a friend or foe detection must be implemented. If multiple enemies are near only one should be attacked per round/tick.

There also need to be enemies on the map. Therefore the currenlty unknown wandering elements should be converted to enemies.

#62 is related for the detection of enemies. But we already have an `Affiliation` so that is not a requirement.
#65 is related for the stats of elements. The blueprints may already provide defaults/inital values.

Loomie:
Elements have 100 health and 20 fire power. Nearby enemies will be attacked each round.


Loomie:
I implemented the data class `CombatStats` in the model beside the elements. The behavior `Fighting` checks the neighbor coordinates of the current position and asks the context if it has entities there. The ElementEntities are added and removed from the Game based on the alive state which is derived from the health: If it is above zero the element is alive and else dead.

The drawing is not so great: If a dead unit moves it is shown again because of its position change 🙄 Maybe that's a race condition between the wandering and the fighting behavior!?

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/Game.kt
```diff
@@ -13,19 +16,40 @@ import kotlin.random.Random
...
     /**
      * Adds the given entity to the engine and returns it to make chained calls possible.
      */
-    fun <T : EntityType> addEntity(entity: GameEntity<T>) = entity.also { engine.addEntity(it) }
+    fun <T : EntityType> addEntity(entity: GameEntity<T>) = entity.also {
+        engine.addEntity(it)
+        if (it.type is ElementType) {
+            val element = it as ElementEntity
+            allElements.add(element)
+            // TODO does this logic belong here?
```
Baret, line 38: I think this is exactly the spot where this belongs. The `Game` is managing all entities that are present in... well, the game. It creates new ones and adds them to the engine. And I think this is an elegant way to get rid of "dead" entities.
:+1:

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/entities/types/Combatant.kt
```diff
@@ -0,0 +1,23 @@
...
+typealias CombatantEntity = GameEntity<Combatant>
+
+/** Access to [CombatStats] */
+val CombatantEntity.combatStats: CombatStats
+    get() = getAttribute(CombatAttribute::class).stats
+
+/** This combatant attacks the given [target] */
+internal fun CombatantEntity.attack(target: CombatantEntity) {
+    combatStats.attack(target.combatStats)
+}
```
Baret, line 23:
```suggestion
internal infix fun CombatantEntity.attack(target: CombatantEntity) = combatStats.attack(target.combatStats)
```

Then you can simply do `attacker attack enemyToAttack` in `Fighting` without all that "syntactic noise".

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/behaviours/Fighting.kt
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    fun attackNearbyEnemies(attacker: ElementEntity, context: GameContext) {
+        val currentPosition = attacker.currentPosition
+        val enemyToAttack = currentPosition.neighbors()
+                .map(context::findElementAt)
+                .filter { it.isPresent && it.get().affiliation == Affiliation.Hostile }
+                .map(Maybe<ElementEntity>::get)
+                .firstOrNull()
+        if (enemyToAttack != null) {
```
Baret, line 34: I feel this could be written more fluent. Either by replacing the if with `?.` or by not mapping to `Maybe::get` but just finding the fist `Maybe` and calling `ifPresent`

game/engine/src/test/kotlin/de/gleex/pltcmd/game/engine/systems/behaviours/FightingTest.kt
```diff
@@ -0,0 +1,71 @@
...
+    val context = mockk<GameContext>()
+    every { context.findElementAt(any()) } returns Maybe.empty()
+    every { context.findElementAt(targetPosition) } returns Maybe.of(target)
+    return Pair(attacker, context)
+}
+
+private fun assertCombatResult(attackerStats: CombatStats, targetStats: CombatStats, expectedHealth: Int, expectedAlive: Boolean = true) {
+    attackerStats.health.value shouldBe 100
+    targetStats.health.value shouldBe expectedHealth
+    targetStats.isAlive.value shouldBe expectedAlive
```
Baret, line 70:
```suggestion
    assertSoftly {
        attackerStats.health.value shouldBe 100
        targetStats.health.value shouldBe expectedHealth
        targetStats.isAlive.value shouldBe expectedAlive
    }
```

game/engine/src/test/kotlin/de/gleex/pltcmd/game/engine/systems/behaviours/FightingTest.kt
```diff
@@ -0,0 +1,71 @@
...
+import io.mockk.mockkStatic
+import org.hexworks.cobalt.datatypes.Maybe
+
+class FightingTest : StringSpec({
+    // enable mocking of extensions
+    mockkStatic("de.gleex.pltcmd.game.engine.entities.types.CombatantKt")
+    mockkStatic("de.gleex.pltcmd.game.engine.entities.types.ElementTypeKt")
+    mockkStatic("de.gleex.pltcmd.game.engine.entities.types.PositionableKt")
+
+    "attackNearbyEnemies with single enemy" {
```
Baret, line 22: This looks like a good test setup. But when you are already at it, I would like to see a test for multiple enemies, too. Something simple like "repeat(5) { attack enemy 1 }; assert enemy1 down; assert enemy2.health = 100" and then 5 more attacks and assert again.

As I see it this would need a small refactoring of `setupCombat()`, right? But passing a vararg of targetStats should do it, right?

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/entities/ColorRepository.kt
```diff
@@ -16,6 +16,9 @@ object ColorRepository {
     val FRIENDLY = TileColor.create(42, 42, 254)
     val FRIENDLY_TRANSPARENT = FRIENDLY.withAlpha(ALPHA_BG)
 
+    val HOSTILE = TileColor.create(252, 100, 100)
```
Baret, line 19: The ["original color" in wikipedia](https://upload.wikimedia.org/wikipedia/commons/e/ee/NATO_Map_Symbol_-_Hostile_Unit_%28Old%29.svg) is "#fb0421" which translates to `251, 4, 33` ;)

But after all we will surely need to revamp the tiles, their colors and probably more sonner or later.

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/combat/CombatStats.kt
```diff
@@ -0,0 +1,26 @@
...
+ * All values needed for combat.
+ * @param firepower the amount of damage done per tick
+ * @param health the current amount of health left
+ */
+data class CombatStats(val firepower: ObservableValue<Int> = 20.toProperty(), val health: Property<Int> = 100.toProperty()) {
+    /** return false if the health reaches 0 */
+    val isAlive: ObservableValue<Boolean> = health.bindTransform { it > 0 }
+
+    fun attack(target: CombatStats) {
+        target.health.transformValue { it - firepower.value }
```
Baret, line 19: Oh, didn't notice `transformValue` yet. What's the difference to a simple `updateValue()`? There you would need to call `healt.value` again, right?

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/attributes/CombatAttribute.kt
```diff
@@ -0,0 +1,7 @@
+package de.gleex.pltcmd.game.engine.attributes
+
+import de.gleex.pltcmd.model.elements.combat.CombatStats
+import org.hexworks.amethyst.api.Attribute
+
+/** The [CombatStats] of a fighting entity. */
+internal class CombatAttribute(val stats: CombatStats = CombatStats()) : Attribute
```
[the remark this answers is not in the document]
Loomie: Breaks iterating the attributes (I guess) -> https://github.com/Kotlin/kotlinx.collections.immutable/issues/76

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/behaviours/Fighting.kt
Loomie: I also thought that. But its not possible: `?.` cannot be used because the parameter may be null not the object on which the method is called. `Maybe` cannot be used because the list of neighbors may be empty and therefore no maybe is found (theoretically) or else there is a Maybe that contains a Maybe.
To reverse the caller and parameter an extension function may be used: ElementEntity.getAttackedBy(other) but sounds strange. Maybe I'm missing something here 🤔

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/entities/ColorRepository.kt
Loomie: Didn't know about the "orignal colors".

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/combat/CombatStats.kt
Loomie: Yes. I have to get the current value, substract the damage and set the new value. `transform` gives me the current value so I don't need another call for it like `health.update(health.value - firepower)`.
But know there is a call the current value anyways to check if it is already at zero.
What brings to: If the damage is more than the health should it be negative or exactly zero?

Loomie:
Done what was requested.

model/elements/src/main/kotlin/de/gleex/pltcmd/model/elements/combat/CombatStats.kt
Baret: It think that doesn't matter in the end. The important part is that the alive boolean checks for "> 0" or the death check for "<= 0" (instead of "== 0") respectively. When the HP would be visible somewhere it should never be displayed as negative. But in that case it is irrelevant if the UI just displays it as 0 or if the `health`-Property has a validator directly.

Baret:
Now you just have to resolve the conflicts :)

PR #92, opened 2020-07-26
resolves #90: Commander's intent

### LINKED ISSUE #90: Commander
**Basic idea/big picture**
What is your idea? What  new feature would you like to see?

**Details**
What are the details? Is there anything that must not be missing?

**Technical background**
Do already know any technical details, for example where to put it or how to implement the feature?

Are there other issues related to this feature? What has to be done before this feature can be implemented?

Baret:
PR for #90


Baret:
I now have the basic infrastructure for `Goal`s.

To have something to work with I first reworked the movement system. A `MovableEntity` now has a `baseSpeedInKph` and a `MovementPath` (determined by the `PathFinding` facet), the end of which is the current destination (so the `DestinationAttribute` is obsolete). And then there are `MovementModifiers` that modify the base speed which results in the actual `currentSpeed`. These modifiers can either cap the speed to a max, modify it or completely prevent movement (i.e. pinned, halted etc.).

With that let's come to the goals, the actual point of this PR :D
The basic idea is that the systems of an element represent the very basics an element can do, which is basically "move and shoot". Maybe more later.
On top of that are the goals or rather the one current goal that might use smaller sub-goals to be completed. I think you can imagine it like follows: The HQ gives the commander of an element (let's say a Platoon-Commander) an order. This order is the current "commander's intent" that the plt-commander tries to fulfil. It is set as the current goal of the element. This goal is kind of the brain of the plt-commander. He decides what needs to be done to complete the commander's intent. It contains the more abstract logic of what the element needs to do with its rudimentary abilities. He orders his soldiers to move somewhere, to halt or to engage an enemy. But he also adjusts his plan which means he updates his current goal. For example the element might come under fire, then the current goal is moved behind the goal "eliminate the current threat". When this is finished, the former goal is continued.

I sense that it may happen that we do not exactly know where to draw the line between "rudimentaries of an element" represented by `Facets` and `Behaviors` and "abstract goals" on top of these. You could probably also implement all these goals with facets and behaviors but I think that would make the system lists of an entity vastly confusing. This way you have "the basics" (inclusive the "execute a goal"-behavior") in facets and behaviors and many small goal-building-blocks in the `Goal` implementations. I believe that makes building more goals much esier as you do not need to extend the entity creation when doing so. You simlpy create a new goal, tell it which sub-goals to execute and you are done.

*What is the current state*
As I mentioned first I have the very basics. I can set a "reach a destination" goal, it gets executed, it can be paused by security halts or "Halt"-Orders from HQ. This means you can have a goal and set a "but now do _this_ first" or you can do a "do _this_ in _x_ turns".
Which is missing now for me to make this system (for now) complete is that goals now their "parent goal" (the one that started it) so they can fail and tell their parent goal about it.
This will require pretty much rework I guess, but I will see :D

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/attributes/goals/Goal.kt
```diff
@@ -0,0 +1,101 @@
...
+import org.hexworks.cobalt.logging.api.LoggerFactory
+import java.util.*
+
+/**
+ * An element may have a _goal_ that is an abstraction layer on top of the basic capabilities of "a bunch of soldiers".
+ *
+ * It gets [step]ped by a behavior until it [isFinished]. Goals may contain sub-goals so each goal can represent any
+ * level of abstraction until it breaks down to a [Command] that needs to be executed by the entity at the lowest level.
+ */
+abstract class Goal(vararg subGoals: Goal) {
```
Baret, line 16: The warning shown here by IntelliJ is probably fixed in Kotlin 1.4: https://youtrack.jetbrains.com/issue/KT-36160

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/facets/ExecuteOrder.kt
```diff
@@ -0,0 +1,65 @@
...
+
+    override suspend fun executeCommand(command: Command<out EntityType, GameContext>) =
+            command.responseWhenCommandIs(OrderCommand::class) { (order, orderedBy, orderedTo, context, entity) ->
+                runBlocking {
+                    when (order) {
+                        MoveTo        -> {
+                            log.debug("Sending MoveTo command for destination $orderedTo")
+                            entity.commandersIntent.set(
+                                    ReachDestination(orderedTo!!),
+                                    RadioGoal(Conversations.Messages.destinationReached(entity.callsign, orderedBy)))
```
Baret, line 34: I would rather liked an API like
```kotlin
entity.commandersIntent(ReachDestination)
  .andThen(RadioGoal)
  .andThen(...)
```

IMO this would be more readable (especially if the goal-chains get longer later, who knows) but in that case I couldn't simply use `goalStack.push()` as the "andThen"-goal would need to be put on the bottom of the stack. Do you think this is readable enough?
Maybe I will simply rework it if I find the time :P

model/radio/src/test/kotlin/de/gleex/pltcmd/model/radio/testhelpers/Matchers.kt
```diff
@@ -1,35 +1,7 @@
 package de.gleex.pltcmd.model.radio.testhelpers
 
 import de.gleex.pltcmd.model.radio.broadcasting.SignalStrength
-import io.kotest.matchers.Matcher
-import io.kotest.matchers.MatcherResult
 import io.kotest.matchers.doubles.shouldBeExactly
-import io.kotest.matchers.should
-import org.hexworks.cobalt.datatypes.Maybe
-
-// - - - Matchers for Maybe
```
Baret, line 10: They were left over here as we copy&pasted them over to the testutils module (and forgot them here).

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/behaviours/MovingForOneMinute.kt
```diff
@@ -0,0 +1,60 @@
...
+                PositionAttribute::class,
+                MovementSpeed::class,
+                MovementPath::class,
+                MovementProgress::class
+        ) {
+
+    private val log = LoggerFactory.getLogger(MovingForOneMinute::class)
+
+    @Suppress("UNCHECKED_CAST")
+    override suspend fun update(entity: Entity<EntityType, GameContext>, context: GameContext): Boolean {
```
Loomie, line 31: previously it was `entity: AnyGameEntity`

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/behaviours/MovingForOneMinute.kt
```diff
@@ -0,0 +1,60 @@
...
+        entity as MovableEntity
```
Loomie, line 32: This is redundant as ElementType is Movable

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/behaviours/MovingForOneMinute.kt
```diff
@@ -0,0 +1,60 @@
...
+        entity as ElementEntity
+        if(entity.canMove.not()) {
```
Loomie, line 34: Please change the property to `canNotMove`.

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/behaviours/MovingForOneMinute.kt
```diff
@@ -0,0 +1,60 @@
...
+            log.debug("${entity.callsign} can not move currently! The current speed is even ${entity.currentSpeedInKph} km/h")
+            return false
+        }
+        return if(entity.movementPath.isNotEmpty()) {
+            val travelDistanceInTiles = entity.currentSpeedInKph / 6.0
```
Loomie, line 39: What is 6.0? I guess it has something to do with 60 minutes of an hour 🤔 As the result is a distance in tiles it seems that it must be of tiles * h/km.

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/behaviours/MovingForOneMinute.kt
```diff
@@ -0,0 +1,60 @@
...
+            entity.movementProgress += travelDistanceInTiles
+            log.debug("${entity.callsign} travels $travelDistanceInTiles tiles/tick with a base speed of ${entity.baseSpeedInKph} and current speed of ${entity.currentSpeedInKph} km/h. New progress: ${entity.movementProgress}")
+            while (entity.movementProgress >= 1.0 && entity.movementPath.isNotEmpty()) {
+                val oldPosition = entity.position.value
+                val newPosition = entity.movementPath.pop()
+                entity.executeCommand(UpdatePosition(oldPosition, newPosition, context, entity))
```
Loomie, line 45: check return value?

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/behaviours/MovingForOneMinute.kt
```diff
@@ -0,0 +1,60 @@
...
+                if(oldPosition != entity.position.value) {
+                    log.debug("${entity.callsign} successfully moved from $oldPosition to ${entity.position}")
+                    entity.movementProgress -= 1.0
```
Loomie, line 48: the name is not so great. First I thought `progress` is a percentage from 0.0 to 1.0. But it is `remainingTravelDistanceUntilEndOfMinute`.

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/behaviours/MovingForOneMinute.kt
```diff
@@ -0,0 +1,60 @@
...
+        entity as ElementEntity
+        if(entity.canMove.not()) {
+            log.debug("${entity.callsign} can not move currently! The current speed is even ${entity.currentSpeedInKph} km/h")
+            return false
+        }
+        return if(entity.movementPath.isNotEmpty()) {
+            val travelDistanceInTiles = entity.currentSpeedInKph / 6.0
+            entity.movementProgress += travelDistanceInTiles
+            log.debug("${entity.callsign} travels $travelDistanceInTiles tiles/tick with a base speed of ${entity.baseSpeedInKph} and current speed of ${entity.currentSpeedInKph} km/h. New progress: ${entity.movementProgress}")
+            while (entity.movementProgress >= 1.0 && entity.movementPath.isNotEmpty()) {
```
Loomie, line 42: The 1.0 should be a constant instead of duplicating the value below for the calculation.

game/options/src/main/kotlin/de/gleex/pltcmd/game/options/GameOptions.kt
```diff
@@ -19,11 +20,28 @@ object GameOptions {
...
+     */
+    object tickRate {
+        val duration: Long = 1L
+        val timeUnit: TimeUnit = TimeUnit.SECONDS
+    }
+
     /**
      * Seed used to always create the same map which is useful for development.
      */
     const val DEBUG_MAP_SEED: Long = 5488540751658L
```
Loomie, line 34: should be private now

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/behaviours/MovingForOneMinute.kt
Baret: I thought about that and I do not know what to check for. I let the entity update its position. What I want to know afterwards is "did it actually update its position?".
When I check the `Response` of `executeCommand()` I do not know what it means. `Pass` most probably means that nothing happened. A `CommandResponse` would be ignored as it simply did not get picked up. And would also not tell me if anything happened. A `Consumed` _may_ mean that a system executed the command and updated the position but it might as well have vetoed against it and consumed it before the `PositionChanging`-facet could do its job.

game/application/src/main/kotlin/de/gleex/pltcmd/game/application/main.kt
```diff
@@ -24,6 +32,9 @@ import org.hexworks.zircon.api.screen.Screen
 import java.util.concurrent.TimeUnit
 import kotlin.random.Random
 
+private val log = LoggerFactory.getLogger("main")
```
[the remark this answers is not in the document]
Baret: There is no class... o_o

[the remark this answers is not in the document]
Loomie: Use "de.gleex.pltcmd.game.application.MainKt". I don't know where IDEA gets this from but it is a reference for the code container as in `de.gleex.pltcmd.game.application.MainKt#main`.

After reading some [documentation](https://kotlinlang.org/docs/reference/reflection.html#property-references)
> To access properties as first-class objects in Kotlin, we can also use the :: operator:
I found this code:
```kotlin
::main::class
```
It gives (with `toString()`) "class de.gleex.pltcmd.game.application.MainKt$main$1"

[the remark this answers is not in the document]
Baret: Looks good! 👍 
It gives me
```
(...) DEBUG d.g.p.game.application.MainKt$log$1 - Bravo is a Fighting Infantry Platoon Bravo with a speed of 3.5 kph.
```

Baret:
I resolved or answered all your comments. Have another review please. I will resolve the conflicts afterwards.

Baret:
I changed the internal data structure for the sub-goals to a deque instead of a stack but externally treat it like a stack (with bonus features ^^).
Now the API of `CommandersIntent` also looks nicer:
```kotlin
entity.commandersIntent
        .setTo(ReachDestination(orderedTo!!))
        .andThen(RadioGoal(Conversations.Messages.destinationReached(entity.callsign, orderedBy)))
        .andThen(PatrolAreaGoal(CoordinateRectangle(orderedTo.movedBy(-5, -5), 10, 10)))
```

PR #95, opened 2020-08-09
Resolves #75: New info sidebar

### LINKED ISSUE #75: Information/Status sidebar
**Basic idea/big picture**
The current window size allows for a wider aspect ratio as most screen resolutions are 16:10 or 16:9. We can use the additional space with another sidebar on the right.
My idea is to use it for context sensitive information. On the map there will be elements, bases, manually put markers and maybe much other stuff. The user needs a way to find out about that. So the right sidebar shows everything there is in the tile the mouse cursor hovers over. This could either be just the top thing you see or a structured list of everything in that block from the bottom up (terrain, unit(s), overlay). Something like that:

```
------------------------
| Terrain: Forest, 8   |
|                      |
| Marker: LZ           |
|                      |
| Elements:            |
|  - Alpha             |
|     Friendly         |
|     Infantry         |
|  - Falcon            |
|     Friendly         |
|     AerialTransport  |
|                      |
------------------------
```

**Details**
The user should have the ability to click a tile to toggle it as selected so the information won't change when hovering over other tiles.

Baret:
PR for #75


Baret:
I restructured the UI and also the packages in the ui module. I tried to make some basic groundwork for #109 in the direction of having `View -> Components -> Fragments`. The `CustomComponent` is a workaround for now, maybe we find a better solution for that in #109.
The new layout actually feels quite well imo. For me this is the maximum the window may grow as I take my laptop resolution (1920x1080, which is a reasonalbe lower bound for resolutions I'd say) as reference. In later zircon versions we might even have a nice fullscreen version.

Please have a look at it.

game/application/src/main/resources/logback.xml
```diff
@@ -1,16 +1,43 @@
...
+        </encoder>
+    </appender>
+
     <logger name="org.hexworks.zircon" level="WARN"/>
     <logger name="org.hexworks.cobalt" level="WARN"/>
-<!--    <logger name="org.hexworks.amethyst.api.entity.Entity" level="DEBUG"/>-->
+    <!--    <logger name="org.hexworks.amethyst.api.entity.Entity" level="DEBUG"/>-->
 
     <logger name="de.gleex.pltcmd.game.application" level="DEBUG"/>
+    <logger name="de.gleex.pltcmd.game.engine.systems" level="INFO"/>
```
Loomie, line 36: The level INFO is already given at the root level. If you want this for debugging purposes set it to debug and comment it out.

game/application/src/main/resources/logback.xml
```diff
@@ -1,16 +1,43 @@
...
+
+    <logger name="de.gleex.pltcmd.game.engine.Game" level="TRACE">
```
Loomie, line 38: Use [additivity="false"](http://logback.qos.ch/manual/configuration.html#overrridingCumulativity) to not have all the TRACE messages in the STDOUT.

game/application/src/main/resources/logback.xml
```diff
@@ -1,16 +1,43 @@
 <configuration scan="true" scanPeriod="30 seconds">
 
+    <jmxConfigurator/>
+    <filter class="ch.qos.logback.access.filter.CountingFilter">
+        <name>countingFilter</name>
```
Loomie, line 5: Unused?

game/ui-strings/src/main/kotlin/de/gleex/pltcmd/game/ui/strings/transformations/Terrain.kt
```diff
@@ -0,0 +1,69 @@
...
+}
+
+internal val terrainTypeTransformation: Transformation<TerrainType> = { format ->
+    when(format) {
+        Format.ICON    -> when (this) {
+            TerrainType.GRASSLAND     -> "G"
+            TerrainType.FOREST        -> "F"
+            TerrainType.MOUNTAIN      -> "M"
+            TerrainType.HILL          -> "H"
+            TerrainType.WATER_DEEP    -> "D"
```
Loomie, line 27: W for water is better.

game/ui-strings/src/main/kotlin/de/gleex/pltcmd/game/ui/strings/transformations/Terrain.kt
```diff
@@ -0,0 +1,69 @@
...
+            TerrainType.WATER_DEEP    -> "Wdp"
+            TerrainType.WATER_SHALLOW -> "Wsh"
+        }
+        Format.SHORT5  -> when (this) {
+            TerrainType.GRASSLAND     -> "Grass"
+            TerrainType.FOREST        -> "Fores"
+            TerrainType.MOUNTAIN      -> "Mount"
+            TerrainType.HILL          -> "Hills"
+            TerrainType.WATER_DEEP    -> "WDeep"
+            TerrainType.WATER_SHALLOW -> "WShlw"
```
Loomie, line 44: "DpWtr" and "ShWtr"?

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/components/InfoSidebar.kt
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * The info sidebar displays contextual information for the player like "what is on that tile?".
+ */
+class InfoSidebar(height: Int, map: GameComponent<*, *>, gameWorld: GameWorld, game: Game) : Fragment {
+
+    private val timePanel = Components.panel()
+            .withSize(UiOptions.SIDEBAR_WIDTH, TickFragment.FRAGMENT_HEIGHT + 2)
+            .withDecorations(ComponentDecorations.box(BoxType.LEFT_RIGHT_DOUBLE))
+            .build()
```
Loomie, line 32: Why is the `TickFragment` not directly added? It would be more consistent because here is also the size set. Below this size is again retrieved as `contentSize.width` - that is exactly the constant `UiOptions.SIDEBAR_WIDTH`, right? If put here it will be more clear (at least for me).

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/components/InfoSidebar.kt
```diff
@@ -0,0 +1,112 @@
...
+import org.hexworks.zircon.api.uievent.UIEventPhase
+
+/**
+ * The info sidebar displays contextual information for the player like "what is on that tile?".
+ */
+class InfoSidebar(height: Int, map: GameComponent<*, *>, gameWorld: GameWorld, game: Game) : Fragment {
+
+    private val timePanel = Components.panel()
+            .withSize(UiOptions.SIDEBAR_WIDTH, TickFragment.FRAGMENT_HEIGHT + 2)
+            .withDecorations(ComponentDecorations.box(BoxType.LEFT_RIGHT_DOUBLE))
```
Loomie, line 31: Why has the time panel another kind of border? I think the the same titled border would be more consistent. Although I noticed that different borders only after seeing this code ;D

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/components/InfoSidebar.kt
```diff
@@ -0,0 +1,112 @@
...
+            .build()
+
+    private val intelPanel = Components.vbox()
+            .withSize(UiOptions.SIDEBAR_WIDTH, height - timePanel.height)
+            .withSpacing(1)
+            .withDecorations(ComponentDecorations.box(BoxType.DOUBLE, "Intel"))
+            .build()
+
+    override val root =
```
Loomie, line 40: "root" is the top most element of a tree. But here it is somewhere in the middle of the UI elements tree. I would name it "sidebar" or just "container".

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/components/InfoSidebar.kt
```diff
@@ -0,0 +1,112 @@
...
+        val fragmentWidth = intelPanel.contentSize.width
+
+        val fragments: List<TileInformationFragment> = listOf(
+                CurrentCoordinateFragment(fragmentWidth, observedTile),
+                TerrainDetailsFragment(fragmentWidth, observedTile, gameWorld.worldMap),
+                MarkersFragment(fragmentWidth, observedTile),
+                ElementInfoFragment(fragmentWidth, observedTile, game)
+        )
+
+        val lockedState: Property<Boolean> = false.toProperty()
```
Loomie, line 70: Add a comment what this lock state means and why you connect it to the fragments.

Loomie: Or put it in a self explanatory function like `lockIntelOnMapMouseRightClick()`.

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/components/InputSidebar.kt
```diff
@@ -0,0 +1,57 @@
...
+                    .build()
+
+    init {
+        val sidebarWidth = root.contentSize.width
+
+        val mapOffset = map.absolutePosition.toProperty()
+        val commandFragment = ElementCommandFragment(sidebarWidth, gameWorld, commandingElement, elementsToCommand, mapOffset, game)
+        root.addFragment(commandFragment)
+        map.handleMouseEvents(MouseEventType.MOUSE_CLICKED) { event, phase ->
+            mapOffset.updateValue(map.absolutePosition)
```
Loomie, line 47: What is this hacky hack? That was not done [before](https://github.com/Baret/pltcmd/pull/95/files#diff-56624c3d771ec04df26a4a30f1369eccddf1d2781702690324d52f942c7b1b3fL120)...

That the offset of other ui components must be known is ugly enough. But I guess on constructing these objects the other is not yet initalised with its final position. Prior the mouse handler was added after all components where added to the screen. Maybe we can model this with a dedicated function `connectToMap()` which is called in GameView?

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/components/InputSidebar.kt
```diff
@@ -0,0 +1,57 @@
...
+            commandFragment(event, phase)
```
Loomie, line 48: For me this is hard to understand. A `commandFragment.handleMouseEvent()` would be clearer. But when registered directly as handler like previously it would be ok.
```kotlin
map.handleMouseEvents(MouseEventType.MOUSE_CLICKED, commandFragment)
```

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/entities/GameWorld.kt
```diff
@@ -24,7 +24,7 @@ import org.hexworks.zircon.internal.game.impl.TopDownProjectionStrategy
  * - An optional overlay, currently used for debug purposes (radio signal strength), but might later be used by the player to "paint on the map".
  *
  */
-class GameWorld(private val worldMap: WorldMap) :
+class GameWorld(val worldMap: WorldMap) :
```
Loomie, line 27: Can we please reduce this to `internal`? Not everybody should use the internal of the ui!

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/components/InfoSidebar.kt
```diff
@@ -0,0 +1,112 @@
...
+
+    init {
+        val observedTile: Property<Coordinate> = gameWorld.visibleTopLeftCoordinate()
+                .toProperty()
+
+        val fragmentWidth = intelPanel.contentSize.width
+
+        val fragments: List<TileInformationFragment> = listOf(
+                CurrentCoordinateFragment(fragmentWidth, observedTile),
+                TerrainDetailsFragment(fragmentWidth, observedTile, gameWorld.worldMap),
```
Loomie, line 65:
```suggestion
                TerrainDetailsFragment(fragmentWidth, observedTile, game.world),
```

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/fragments/TickFragment.kt
```diff
@@ -24,12 +30,5 @@ class TickFragment(override val width: Int) : BaseFragment {
                                                         bindPlusWith createPropertyFrom(": ")
                                                         bindPlusWith Ticker.currentTimeString)
                                     })
-                                addComponent(Components.
-                                    button().
-                                    withText("TICK!").
-                                    build().
-                                    apply {
-                                        processComponentEvents(ComponentEventType.ACTIVATED) { Ticker.tick() }
```
Loomie, line 32: So the `tick()` method can now be private?!

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/fragments/TickFragment.kt
```diff
@@ -5,12 +5,18 @@ import org.hexworks.cobalt.databinding.api.binding.bindPlusWith
...
+        const val FRAGMENT_HEIGHT = 2
+    }
+
     override val root = Components.
                             vbox().
-                            withSize(width, 3).
+                            withSize(width, FRAGMENT_HEIGHT).
                             build().
                             apply {
                                 addComponent(Components.header().withText("Current Tick"))
```
Loomie, line 22: As mentioned [above](https://github.com/Baret/pltcmd/pull/95/files#r519163738): Put the title in the border decoration as the other panels do.

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/fragments/tileinformation/CurrentCoordinateFragment.kt
```diff
@@ -0,0 +1,35 @@
...
+    override val root = Components.hbox().
+            withSize(width, 1).
+            build().
+            apply {
+                addComponent(Components.
+                        label().
+                        withSize(width, 1).
+                        build().
+                        apply {
+                            withFrontendString(Format.SIDEBAR,
```
Loomie, line 27: This is strange btw. The fragment is given a width so it can be used anywhere. But here you assume that this fragment is used inside a sidebar! Here the `Format` should be chosen dependent on the available width. Maybe as `Format.forWidth(widht)` or as you prefer it: `width.toFormat()`.

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/fragments/tileinformation/CurrentCoordinateFragment.kt
```diff
@@ -0,0 +1,35 @@
...
+                                    "Coordinate: ", currentCoordinate)
+                        })
+            }
+
+    override fun updateInformation(newCoordinate: Coordinate) {
+        currentCoordinate.updateValue(newCoordinate)
```
Loomie, line 33: Why not simply provide an observable `val` that sub classes use all the time? It looks strange that the constructors have an `ObservableValue` but didn't use it. So do the following:
```java
// create an independent Property initialized with the current value
protected val currentCoordinate = currentTile.value.toProperty()
```
If you want subclasses to prevent updating this value create a getter typed with `ObservableValue<Coordinate>` that is backed by this field.

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/fragments/tileinformation/ElementInfoFragment.kt
```diff
@@ -0,0 +1,74 @@
...
+ */
+class ElementInfoFragment(
+        override val width: Int,
+        observedTile: Property<Coordinate>,
+        private val game: Game
+) : TileInformationFragment(observedTile) {
+
+    private val currentElement: Property<Maybe<ElementEntity>> = Maybe.empty<ElementEntity>().toProperty()
+
+    private val callSign: FrontendString<String> = whenElementPresent { it.callsign.toString() }
```
Loomie, line 33:
```suggestion
    private val callSign: FrontendString<String> = whenElementPresent { it.callsign.name }
```

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/views/GameView.kt
```diff
@@ -36,45 +41,52 @@ class GameView(private val gameWorld: GameWorld, tileGrid: TileGrid, private val
 
     companion object {
         private val log = LoggerFactory.getLogger(GameView::class)
+
+        private const val LOG_AREA_HEIGHT = WINDOW_HEIGHT - MAP_VIEW_HEIGHT
+        private const val LOG_AREA_WIDTH = WINDOW_WIDTH
+        private const val SIDEBAR_HEIGHT = WINDOW_HEIGHT - LOG_AREA_HEIGHT
     }
 
     override fun onDock() {
```
Loomie, line 50: You need to update `artwork/screenshots/map.png`.

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/Game.kt
```diff
@@ -85,4 +84,16 @@ data class Game(val engine: Engine<GameContext>, val world: WorldMap, val random
         return addEntity(elementEntity)
     }
 
+    /**
+     * Gets all [ElementEntity]s at the given coordinate.
+     */
+    @ExperimentalTime
```
Loomie, line 90: Why not directly use `@OptIn(ExperimentalTime::class)`. That would reduce the clutter in all the callers.

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/fragments/tileinformation/ElementInfoFragment.kt
```diff
@@ -0,0 +1,74 @@
...
+                        .apply {
+                            withFrontendString(description)
+                        })
+            }
+
+    @ExperimentalTime
+    override fun updateInformation(newCoordinate: Coordinate) {
+        currentElement.updateValue(
+                Maybe.ofNullable(
+                        game.elementsAt(newCoordinate).firstOrNull()))
```
Loomie, line 62: Maybe add a TODO to handle multiple elements?

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/fragments/tileinformation/TerrainDetailsFragment.kt
```diff
@@ -0,0 +1,70 @@
...
+    private val terrainProperty: Property<Terrain> = Terrain.of(TerrainType.GRASSLAND, TerrainHeight.MIN).toProperty()
+
+    private val icon: Icon = Components.icon()
+            .withIcon(TileRepository.createTerrainTile(terrainProperty.value))
+            .build()
+            .apply { iconProperty.updateFrom(
+                    terrainProperty.bindTransform { TileRepository.createTerrainTile(it) })
+            }
+
+    private val fixedText: Label = Components.label()
```
Loomie, line 42: Not a header?

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/fragments/tileinformation/TerrainDetailsFragment.kt
```diff
@@ -0,0 +1,70 @@
...
+            .withText(FILLER_TEXT)
+            .withSize(FILLER_TEXT.length, 1)
+            .build()
+
+    private val terrainDesciption: Label = Components.label()
+            .withSize(width - icon.size.width - fixedText.size.width, 1)
+            .build()
+            .apply {
+                withFrontendString(Format.SIDEBAR, terrainProperty)
```
Loomie, line 51: Here is much less space than `Format.SIDEBAR` as the fixed text and icon already consume chars!

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/views/GameView.kt
```diff
@@ -36,45 +41,52 @@ class GameView(private val gameWorld: GameWorld, tileGrid: TileGrid, private val
...
-                build().apply {
+
+        val map = GameComponents.newGameComponentBuilder<Tile, GameBlock>()
+                .withGameArea(gameWorld)
+                .withSize(gameWorld.visibleSize.to2DSize())
+                .build()
+
+        val leftSidebar = InputSidebar(SIDEBAR_HEIGHT, game, commandingElement, elementsToCommand, map, gameWorld)
+
+        val logArea = Components.logArea()
```
Loomie, line 59: Put this first as it is first in the ui.

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/views/GameView.kt
```diff
@@ -110,25 +122,6 @@ class GameView(private val gameWorld: GameWorld, tileGrid: TileGrid, private val
...
-        sidebar.addFragment(TickFragment(sidebarWidth))
-
-        if (GameOptions.displayRadioSignals.value) {
-            val radioSignalFragment = RadioSignalFragment(sidebarWidth)
-            map.handleMouseEvents(MouseEventType.MOUSE_CLICKED, RadioSignalVisualizer(gameWorld, radioSignalFragment.selectedStrength, radioSignalFragment.selectedRange, map.absolutePosition))
-            sidebar.addFragment(radioSignalFragment)
-        }
-
-        sidebar.addFragment(ThemeSelectorFragment(sidebarWidth, screen))
-        sidebar.addFragment(TilesetSelectorFragment(sidebarWidth, map, sidebar))
```
Loomie, line 131: Delete those classes?

Loomie (commented):
I think we should define what should be highlighted bold in the info panel as it is currently not very consistent. Either all titles should be bold as a header. Or the important information like the coordinate is highlighted to be easily readable.

For the code see comments.

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/entities/GameWorld.kt
Loomie: Done in 4c4b75eb

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/fragments/tileinformation/CurrentCoordinateFragment.kt
Loomie: Done in 8a6c69e0

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/views/GameView.kt
Loomie: Done in 161a538b

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/components/InputSidebar.kt
Loomie: Done in 0e1c9c94

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/views/GameView.kt
Loomie: Done in 066826ef

game/application/src/main/resources/logback.xml
Baret: Yup, does not work :(

game/ui-strings/src/main/kotlin/de/gleex/pltcmd/game/ui/strings/transformations/Terrain.kt
Baret: I wanted to have a common beginning so it is alphabetically sorted close to each other. But probably "Wtr" is a better abbreviation than "W".

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/components/InfoSidebar.kt
Baret: Well then it is subtle enough ;D
I don't know, maybe I wanted to highlight it a bit. I can play around with it and see what looks best.

Baret: Ok I tried giving both panels a DOUBLE box and it looks aweful. So I rather went for LEFT_RIGHT_DOUBLE for both panels (intel and time). Looks nice imo

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/components/InfoSidebar.kt
Baret: It is an `override` and is the `Component` a `Fragment` resembles around. It is needed and can not have a different name.

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/components/InputSidebar.kt
Baret: Thank you so much! Yes, there was a circular dependency so that the map did not yet have an absolute position when I had to give it to the sidebar. And yes it was a hacky hack to put it into a property just to update it once.
I like you `connectTo(map)` approach.
But because of this I opened [this issue](https://github.com/Hexworks/zircon/issues/335), so maybe we can get rid of that absolute position stuff at some time.

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/fragments/TickFragment.kt
Baret: Indeed 👍

Baret: That kind of led me to quickly make the game pausable ^^
You can now pause/unpause with P or spacebar, couldn't decide which binding would be more useful and intuitive and as spacebar is not yet used, I took both (for now).

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/fragments/tileinformation/CurrentCoordinateFragment.kt
Baret: The behavior of `withFrontendString()` still has some flaws anyways. Currently it converts all given objects to a `FrontendString` with the given format (here SIDEBAR). This means that they do not get correctly truncated when their combined length becomes longer that SIDEBAR.
I would like to have two much better behaviors:
First: a version of `withFrontendStrings` which takes a "maximum length format", like SIDEBAR in this example, and only a list of `FrontendStrings`. Those already have a format that should stay. This means you can append 3 SHORT5 string together and they would only use 15 characters while each transforming to the correct 5-char-string. Currently the first one would be cornverted to SIDEBAR and might thus take up all available space.
Second, your idea: For a given width find the largest `Format` possible. Or a simple `Format.FLEXIBLE(length)` for "I want a fixed string here".

I will open up a new issue for that though.

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/fragments/tileinformation/TerrainDetailsFragment.kt
Baret: Yes, see my comment [here](https://github.com/Baret/pltcmd/pull/95#discussion_r526060840).

I "just know" that there is enough space to display the whole terrain name. I also could've used `Format.FULL`. That is exactly why I want a more fine grained behavior here, but that will be done in another issue.

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/views/GameView.kt
Baret: Good idea as they are also provided by Zircon anyways.

Baret:
> I think we should define what should be highlighted bold in the info panel as it is currently not very consistent. Either all titles should be bold as a header. Or the important information like the coordinate is highlighted to be easily readable.

Yes, I will redesign the sidebar a bit. It probably looks best to have the titles be bold and have the content in a new line, a bit indented like so:
```
Coordinate:
  (123|234)

Markers:
  none

Element(s)
  - Alpha
    (...)
```

PR #111, opened 2020-08-25
Resolves #87: frontend strings

### LINKED ISSUE #87: Frontend strings
**Basic idea/big picture**
As we are using a tile based frontend we will surely often bump into space limitations when displaying things. For example the sidebar(s) has only  a specific width. Dialogs or other future UI elements will probably also have some size that can only display that much text. For that I imagine a unified way of representing objects as strings in different locations.

**Details**
It is probably best to define some default widths to always know how much space there is to display text. Fragments that are only displayed in the sidebar are always of width 37 for example (according to [the current config](https://github.com/Baret/pltcmd/blob/master/game/options/src/main/kotlin/de/gleex/pltcmd/game/options/UiOptions.kt)). The "shop" might display elements in a box of fixed width and height which would be another standard.

**Technical background**
I imagine some sort of `FrontendString` that internally handles cutting an original base string conveniently into the right sizes. It's probably best to just lay out my idea of how the class/interface looks like.

```kotlin
class FrontendString(val original: String) {
    fun forLength(i: Int): String // Default behaviour for any given length. Might not need to be public but just used internally
    
    val sidebar: String // i.e. calls forLength(37)
    val icon: Tile // for when there is only a single tile, basically length = 1
    
    // ... probably more vals here

    // there should also be ways to get zircon specific StringComposites. Maybe double all vals so you can get String and StringComposite?
}
```

To not clutter the actual classes with frontend stuff this should reside in an own module as maybe many model classes might need to be converted to user readable text. A frontend string is created by calling an extension funtion on the object. The default is

```kotlin
fun Any.toFrontendString(): FrontendString
```

Which creates a default implementation. It just uses `.toString()` as original string and abbreviates it with "..." when it gets too long. But this mechanism might be handy to easily create reasonable short variants of units or elements. For example the unit "HMGTeam" might be abbreviated to "HMG", "Main battle tank" to "MBT" and so forth.

By creating `StringComposite`s (I am not sure about the exact name) coloring of text is also possible.

Baret:
PR for issue #87


Baret:
I created a new module `ui-strings`. First I wanted to put it under `ui` but that did not work without it having a packaging of `pom` so I put the new module "next to it". Maybe we need a `ui` parent and respective groupID in the future....

The module contains a very basic interface `FrontendString` and a basic default implementation. The only specialization so far is the `CoordinateFrontendString` which is being used in the mouse position fragment.
More specializations I see coming are for `Unit` and `Terrain` at least. And maybe we will need something like `asIcon(): Icon` or `asLabel(): Label` to directly create `Component`s from it. And _then_ it might be necessary to put the `TileRepository` in an own module (you could simply use the terrain tile for the `Icon`) which would justify a ui-parent-module.

game/ui-strings/src/test/kotlin/de/gleex/pltcmd/game/ui/strings/special/CoordinateFrontendStringTest.kt
```diff
@@ -0,0 +1,135 @@
...
+    "A coordinate with more than 3 digits" should {
+        val longEasting = Coordinate(1234, 567)
+        Format.values()
+                .forEach { length ->
+                    "use a star for easting when necessary for length $length, testing with coordinate $longEasting" {
+                        val actual = CoordinateFrontendString(longEasting, length).value
+                        when (length) {
+                            Format.ICON    -> actual shouldBe "|"
+                            Format.SHORT3  -> actual shouldBe "*|5"
+                            Format.SHORT5  -> actual shouldBe "(*|5)"
```
Loomie, line 35: Why not using the full `12|05` and omitting the parenthesis? To indicate a main coordinate instead of a normal coordinate we should use another separator like `12!05`.

game/options/src/main/kotlin/de/gleex/pltcmd/game/options/GameOptions.kt
```diff
@@ -49,7 +49,7 @@ object GameOptions {
     /**
      * The origin of the map. In other words the world map's bottom left corner is this coordinate.
      */
-    val MAP_ORIGIN: Coordinate = Coordinate(150, 300)
+    val MAP_ORIGIN: Coordinate = Coordinate(850, 800)
```
Loomie, line 52: Should this be resetted before merging into master?

game/ui-strings/pom.xml
```diff
@@ -0,0 +1,28 @@
...
+
+    <artifactId>ui-strings</artifactId>
+
+    <name>Frontend strings</name>
+    <description>Unified ways to create representations of objects in the frontend.</description>
+
+    <dependencies>
+        <dependency>
+            <groupId>de.gleex.pltcmd.model</groupId>
+            <artifactId>world</artifactId>
```
Loomie, line 20: Do we need one project for each model project to separate the business logic? Like `ui-strings.world`, ui-strings.element` and `ui-strings.radio`. Or will we not collect so much classes in this project? Maybe we just need one or two converters per model project because only very important classes are visible in the ui.

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/fragments/CoordinateAtMousePosition.kt
```diff
@@ -25,12 +28,15 @@ class CoordinateAtMousePosition(override val width: Int, gameComponent: GameComp
                         withSize(width, 1).
                         build().
                         apply {
-                            textProperty.updateFrom(createPropertyFrom("Mouse pos: ") bindPlusWith currentCoordinate.bindTransform { it.toString() }, true)
+                            withFrontendString(
+                                    "Mouse pos: ".toFrontendString()
```
Loomie, line 32: Seems like you also need a `Any.plus(FrontendString)` that internally creates the Observable and FrontendString`.

Loomie (commented):
The implementation looks good. But the structure and external representation may be improved.

game/ui-strings/src/main/kotlin/de/gleex/pltcmd/game/ui/strings/FrontendString.kt
```diff
@@ -0,0 +1,26 @@
...
+import org.hexworks.cobalt.databinding.api.value.ObservableValue
+
+
+/**
+ * A string representation of an object that can be used in the frontend.
+ *
+ * It provides means to trim the string down to various lengths so it will always fit into the currently available
+ * space (i.e. a sidebar or a dialog) while always representing a human readable resemblance of the object.
+ */
+interface FrontendString<out T : Any> : ObservableValue<String> {
```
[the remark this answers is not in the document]
Baret: I wanted to keep the interface as clean as possible. Yes, it is of type `T` and is not directly used or exposed in the interface. I thought about it a lot and tried different things and ended up with this. The interface just says "you have an object of type `T`" and "you are an `Obervable<String>`", so "go make, that you object is transformed to an observed string. But the transformation is not a public thing so I did not want to put something like `transform()` into it. It's nothing a user of a `FrontendString` would call. What is interesting from the public point of view is only that you give it some object and a "format" (basically the length of the resulting string) and that you can bind it to the UI because it is observable.

But I really like your idea of splitting up responsibilities here. The `FrontendString` is simply an observable string, it does not care _how_ an object is transformed into a string.

> The `FrontendString` itself is more a binding than actual data

I thought about that, too. "Do I extend `Observable<String>` or rather `Binding<String>`, but in a binding you have to manage `Subscription`s so I decided to keep it simple :D

game/ui-strings/src/main/kotlin/de/gleex/pltcmd/game/ui/strings/extensions/Arithmetic.kt
```diff
@@ -0,0 +1,23 @@
+package de.gleex.pltcmd.game.ui.strings.extensions
+
+import de.gleex.pltcmd.game.ui.strings.Format
+import de.gleex.pltcmd.game.ui.strings.FrontendString
+import org.hexworks.cobalt.databinding.api.value.ObservableValue
+
```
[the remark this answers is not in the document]
Baret: Yes, the "static" `minOf` could have been in a companion object. Although you have to use `Format.minOf()` then. I have been inspired by the kotlin minOf-functions.

I didn't know how may "calculations" there would come up so I did not want to overload the interface. I figured it made sense to put `operator fun plus(other: FrontendString<Any>): FrontendString<*>` there but converting something to a frontendstring and then calling this `plus` didn't seem like something every implementation has to do.

game/ui-strings/src/test/kotlin/de/gleex/pltcmd/game/ui/strings/special/CoordinateFrontendStringTest.kt
Baret: Great idea having a different separator for main coodinates!

And also good to not waste two digits for the parentheses. But it gets rather complicated to implement it nicely. As I do not exactly know if we will actually use this format for coordinates later I keep it like that for now.

game/options/src/main/kotlin/de/gleex/pltcmd/game/options/GameOptions.kt
Baret: I don't really care. Maybe it is not bad to have a default map with larger coordinates so we don't get too used to only 3 digit northings/eastings.

game/ui-strings/pom.xml
Baret: I would keep it in one project for now and rely on different packages where needed for now. I guess there won't be _that_ many classes that need to be transformed for the UI and I want to make implementing new frontend strings as simple as possible. When you realize "oh, I need to display this object in the frontend" you should simply go to this project and implement something simple like

```kotlin
when(format) {
                Format.ICON                 -> Coordinate.SEPARATOR
                Format.SHORT3               -> objectToTransform.formattedMainCoordinate
                Format.SHORT5               -> "(${objectToTransform.formattedMainCoordinate})"
                Format.SIDEBAR, Format.FULL -> objectToTransform.toString()
            }
```

and that's it. Most of this stuff should be "one liners" ;)

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/fragments/CoordinateAtMousePosition.kt
Baret: I tried... didn't work :(
`String.plus()` seems to win over `Any.plus()`.

Baret:
I figured that a frontend string indeed is just a combination of three things resulting in an observable string: Some object, a _format_ describing how long the string may get and a _transformation_ that transforms the object to said string. Yes, in the end it is basically a `Binding`, something that Cobalt already gives us with `ObservableValue.bindTransform`. But I still wanted to keep it in a unified and more concise manner so I basically wrap it up in the `FrontendString` which now is a class instead of an interface.
All you need to provide, next to the object and the `Format`, is the `Transformation`. Which itself is just a typealias for "transform an object to a string using a format" (`T.(Format) -> String`).

I am satisfied with how it works out for now. It is still open how to extend it if we want to add more advanced stuff like colored strings (i.e. `CharacterTileString`) or create whole `Component`s from observable objects (like a `Label` with a given size/format). It might become as easy as creating extension functins like `TextHolder.withFrontendString`.

BTW. the `Format` enum is not final, of course. We will see which formats we will actually need in the UI.

model/world/src/test/kotlin/de/gleex/pltcmd/model/world/coordinate/MainCoordinateTest.kt
```diff
@@ -24,8 +24,8 @@ class MainCoordinateTest: WordSpec({
         }
     }
 
-    "The string representation of a main coordinate in the form of (x|y)" should {
-        val regex = "\\(-?\\d+\\|-?\\d+\\)"
+    "The string representation of a main coordinate in the form of (x!y)" should {
+        val regex = "\\(-?\\d+\\!-?\\d+\\)"
```
Loomie, line 28: Why not using the constant `SEPARATOR`? Like in `Coordinate.REGEX_STRING`.

game/ui-strings/src/main/kotlin/de/gleex/pltcmd/game/ui/strings/FrontendString.kt
```diff
@@ -1,26 +1,58 @@
...
+import org.hexworks.zircon.api.Components
 
 /**
- * A string representation of an object that can be used in the frontend.
+ * A frontend string is a string representation of any object (i.e. model objects) that is observable and
+ * thus can be bound to UI elements.
  *
- * It provides means to trim the string down to various lengths so it will always fit into the currently available
- * space (i.e. a sidebar or a dialog) while always representing a human readable resemblance of the object.
+ * @sample example
```
Loomie, line 17: I think you wanted to replace this placeholder with a real example?!

game/ui-strings/src/main/kotlin/de/gleex/pltcmd/game/ui/strings/FrontendString.kt
```diff
@@ -1,26 +1,58 @@
...
+    }
+}
 
+private fun example() {
+    Components.label()
+            // ...
+            .build()
+            .withFrontendString(
+                    Coordinate(100, 100)
+                            .toFrontendString(Format.SHORT5))
```
Loomie, line 57: I have to call `withFrontendString` and `toFrontendString()` together? Why is the `toFrontendString()` not done inside of `withFrontednString()`? So just pass the object and format. Or does it then always use the extension function for `Any`?

game/ui-strings/src/main/kotlin/de/gleex/pltcmd/game/ui/strings/FrontendString.kt
```diff
@@ -1,26 +1,58 @@
...
-    /**
-     * The [Format] restricting the space available for this string.
-     */
-    val format: Format
+    private val internalBinding: Binding<String> = originalObject.bindTransform { it.transform(format) }
+
+    override val value: String
+        get() {
+            val transformed = internalBinding.value
+            require(transformed.length <= format.length) {
```
Loomie, line 30: This validation could be put into the `internalBinding` by using `toProperty(validator)`. The type of `internalBinding` can be opened up to `ObservableValue<String>`. An then the Property is your FrontendString ;p ... or you map the value to this value.

Loomie: PS: Maybe you can delegate to that property? Let the inherited value be provided by that delegate. But the `toProperty` creates an `Observable<Binding<String>>` m(
So it seems that the validation is only implemented in BaseProperty. Thus it is not part of the interfaces :/ Looks like you have to do it on your own. And then delegation is not possible as the delegate has to perform the validation on access.

But the value can only be invalid if it changes. So instead of validating each access just the transformation has to be validated!

I implemented that in Commit 74239546

game/ui-strings/src/test/kotlin/de/gleex/pltcmd/game/ui/strings/transformations/DefaultTransformationTest.kt
```diff
@@ -21,5 +22,22 @@ class DefaultTransformationTest : WordSpec({
...
+                            Format.SHORT5  -> actualValue shouldBe "A(foo"
+                            Format.SHORT3  -> actualValue shouldBe "A(f"
+                            Format.ICON    -> actualValue shouldBe "A"
+                        }
+                    }
+        }
     }
 })
+
+private data class A(val foo: Int)
```
Loomie, line 43: Not a very good class name. `SomeDataClass` would be better or something like `WithToString`.

game/application/src/main/kotlin/de/gleex/pltcmd/game/application/examples/units/asStrings.kt
```diff
@@ -0,0 +1,41 @@
...
+
+import de.gleex.pltcmd.game.ui.strings.Format
+import de.gleex.pltcmd.game.ui.strings.extensions.toFrontendString
+import de.gleex.pltcmd.model.elements.units.Units
+
+fun main() {
+    val lineFormat = "| %3d " + Format.values()
+            .joinToString("", "", "|") { it.asFormatString() }
+    val headerLine = "| Idx " + Format.values()
+            .joinToString(" | ", "| ", " |") { it.toFrontendString(it).value }
```
Loomie, line 11: padded to the lenght would be nicer (for SIDEBAR), but problematic for FULL :D

Loomie: Ah you just have to use the `lineFormat` as for all other lines 💡

Loomie: `lineFormat` is not working as it expects the first column to have an integer. But I reused the rest of the format in commit cde5fd22

game/ui-strings/src/main/kotlin/de/gleex/pltcmd/game/ui/strings/extensions/ToFrontendString.kt
```diff
@@ -34,3 +37,14 @@ fun Coordinate.toFrontendString(format: Format = Format.FULL): FrontendString<Co
 fun ObservableValue<Coordinate>.toFrontendString(format: Format = Format.FULL) =
         FrontendString(this, format, coordinateTransformation)
 
+/**
+ * Creates a [FrontendString] of this unit blueprint.
+ */
+fun Units.toFrontendString(format: Format = Format.FULL): FrontendString<Unit> =
+        FrontendString(this.new().toProperty(), format, unitTransformation)
```
Loomie, line 44: I think you can simply use `new().toFrontendString()`.

game/ui-strings/src/main/kotlin/de/gleex/pltcmd/game/ui/strings/transformations/Unit.kt
```diff
@@ -0,0 +1,70 @@
...
+    }
+}
+
+/**
+ * A simple helper method to have minimal syntax overhead for each unit.
+ *
+ * All string representations except the full string are optional. For every format that is not provided [full] will
+ * be formatted using [defaultTransformation].
+ */
+private fun Format.forUnit(
```
Loomie, line 57: 👍

game/ui-strings/src/main/kotlin/de/gleex/pltcmd/game/ui/strings/transformations/Unit.kt
```diff
@@ -0,0 +1,70 @@
...
+        Units.Officer             -> format.forUnit("Officer", "O", "OFF", "Off.")
+        Units.Radioman            -> format.forUnit("Radioman (RTO)", "P", "RAD")
+        Units.Medic               -> format.forUnit("Medic", "M", "MED")
+        Units.CombatEngineer      -> format.forUnit("Combat engineer", "E", "ENG", "Engi.")
+        Units.HMGTeam             -> format.forUnit("Heavy machine gun team (HMG)", "H", "HMG", "HMG")
+        Units.AntiTankTeam        -> format.forUnit("Anti-tank team", "T", "AT", "ATInf")
+        Units.AntiAirTeam         -> format.forUnit("Anti-air team", "A", "AA", "AAInf")
+        Units.Scout               -> format.forUnit("Infantry scout", "S", "SCI", "Scout")
+        Units.SniperTeam          -> format.forUnit("Sniper team", "X", "SNI", "Snip.")
+        Units.MortarTeam          -> format.forUnit("Mortar team", "I", "MOR", "Mort.")
```
Loomie, line 21: One character is just not unique enought :/

game/ui-strings/src/main/kotlin/de/gleex/pltcmd/game/ui/strings/transformations/Unit.kt
```diff
@@ -0,0 +1,70 @@
...
+import de.gleex.pltcmd.game.ui.strings.Transformation
+import de.gleex.pltcmd.model.elements.units.Unit
+import de.gleex.pltcmd.model.elements.units.Units
+
+internal val unitTransformation: Transformation<Unit> = { format ->
+    when(blueprint) {
+        Units.Rifleman            -> format.forUnit("Rifleman", "R", "RIF", "RifMa")
+        Units.Grenadier           -> format.forUnit("Grenadier", "G", "GRE")
+        Units.Officer             -> format.forUnit("Officer", "O", "OFF", "Off.")
+        Units.Radioman            -> format.forUnit("Radioman (RTO)", "P", "RAD")
```
Loomie, line 13: Why is "RTO" in the full text but it is not used for short3?

game/ui-strings/src/main/kotlin/de/gleex/pltcmd/game/ui/strings/transformations/Unit.kt
```diff
@@ -0,0 +1,70 @@
...
+        Units.Medic               -> format.forUnit("Medic", "M", "MED")
+        Units.CombatEngineer      -> format.forUnit("Combat engineer", "E", "ENG", "Engi.")
+        Units.HMGTeam             -> format.forUnit("Heavy machine gun team (HMG)", "H", "HMG", "HMG")
+        Units.AntiTankTeam        -> format.forUnit("Anti-tank team", "T", "AT", "ATInf")
+        Units.AntiAirTeam         -> format.forUnit("Anti-air team", "A", "AA", "AAInf")
```
Loomie, line 18: It is inconsistent to add 'team' the full version but 'Inf' to the short version. I think we can use "Anti-tank infantry" for the full name?
I see a similar problem for the helicopters: "Transport helicopter" vs. "H-Trn" - in the full version the 'helicopter' is a suffix but the short form has the prefix 'H'.
In general I think the prefix is better to identify the group first and then the corresponding unit type. Like "H-Trn" and "V-Trn" for "Transport helicopter" and "Transport truck" (instead of 'V' for vehicle we can use 'T' for truck). That would also fit for the "Radio truck" which is currently "RadTr" but should be "T-Rad" like the "Radio | Radioman (RTO)" that should be shortened as "I-Rad".

That's how I would think the names should be generated.

game/ui-strings/src/test/kotlin/de/gleex/pltcmd/game/ui/strings/transformations/UnitTransformationTest.kt
```diff
@@ -0,0 +1,72 @@
...
+    init {
+        "The unit transformation" should {
+            "work for units and unit blueprints" {
+                Format.values()
+                        .forAll { format ->
+                            Units.values()
+                                    .forAll { blueprint ->
+                                        val blueprintValue: String = blueprint.toFrontendString(format).value
+                                        val unitValue: String = blueprint.new()
+                                                .toFrontendString(format).value
```
Loomie, line 29: You test `Unit.toFrontendString` with `Unit.toFrontendString`? The expected value is not very fixed!

game/ui-strings/src/test/kotlin/de/gleex/pltcmd/game/ui/strings/transformations/UnitTransformationTest.kt
```diff
@@ -0,0 +1,72 @@
...
+                                    }
+                        }
+            }
+        }
+
+        "The icon for a unit" should {
+            Units.values()
+                    .groupBy { it.kind }
+                    .forEach { (unitKind, blueprints) ->
+                        "be unique for unit kind $unitKind" {
```
Loomie, line 40: Nice! Haven't seen this pattern in the code 🙄

game/ui-strings/src/test/kotlin/de/gleex/pltcmd/game/ui/strings/transformations/UnitTransformationTest.kt
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        "The 3 digit frontend string of all units" should {
+            "be unique" {
+                val short3Strings = Units.values()
+                        .map { it.toFrontendString(Format.SHORT3).value }
+                short3Strings.logDuplicates(Format.SHORT3)
+                short3Strings should beUnique()
+            }
+        }
```
Loomie, line 61: 5 chars should also be unique!?!

Loomie (commented):
Looks pretty good. Awesome work with the unit names! See above comments for further improvements.

game/ui-strings/src/main/kotlin/de/gleex/pltcmd/game/ui/strings/FrontendString.kt
Baret: I thought of the latter. You might already have a local field or property of type `FrontendString` and then the signature `withFrontendString(FrontendString)` would make perfect sense ;)

But I like your idea. Maybe a `withFrontendString(Format, vararg Any)` might be cool. But this probably indeed calls `Any.toFrontendString()` for all objects. I observed that behavior already in `T.toFrontendString()`. It calls `T.toProperty()` which results in an `ObservableValue<T>` but as `T` is of type `Any` and the actual type is erased at runtime `ObservableValue<Any>.toFrontendString()` is called which results in a default transformation. Even if there is a special override, i.e. `ObservableValue<Coordinate>.toFrontendString()` :(
And "[reifying|(https://kotlinlang.org/docs/reference/inline-functions.html#reified-type-parameters)" the type parameter didn't work or help for some reason.

Baret: Yup.. I tried
```kotlin
fun TextHolder.withFrontendString(format: Format, vararg objects: Any) {
    val frontendStrings: Collection<FrontendString<*>> = objects.map {
        when (it) {
            is ObservableValue<Any> -> it.toFrontendString(format) // calls ObservableValue<T>.toFrontendString()
            else                    -> it.toFrontendString(format)  // calls Any.toFrontendString()
        }
    }
//...
}
```

I tried passing SHORT5 and the `Property<Coordinate>` of `CoordinateAtMousePosition`. When not checking the type at runtime with `when` it results in a toString-call of the `Property` which gives "Prope". When smart casting it to an Observable it uses the default Transformation on the observed `Coordinate`.

On my way I already thought of having a map of every type and its transformation (i.e. "Coordinate -> coordTransformation, Unit -> unitTransformation, Any -> defaultTransformation") so I do not need to implement `Something.toFrontendString()` over and over again. That would also solve this problem. But at some point I got stuck and discarded that approach.

game/ui-strings/src/main/kotlin/de/gleex/pltcmd/game/ui/strings/FrontendString.kt
Baret: Nice! I like it! I also tried delegating with `by` along the way but failed ^^

game/ui-strings/src/test/kotlin/de/gleex/pltcmd/game/ui/strings/transformations/DefaultTransformationTest.kt
Baret: I needed a short class name to see more than the class name in the toString representation (i.e. "A(foo" instead of "SomeD").

Actually the property should also be shorter so you can even test for the value with SHORT5...

game/application/src/main/kotlin/de/gleex/pltcmd/game/application/examples/units/asStrings.kt
Baret: nice 👍

game/ui-strings/src/main/kotlin/de/gleex/pltcmd/game/ui/strings/extensions/ToFrontendString.kt
Baret: Yup, maybe I kept it because of readability or something? Dunno :D

But I as long as I only use the blueprint in the transformation I can rather base it on `Units` and don't have to make new instances of units when calling `Units.toFrontendString()` (which currently only exists because of the example code ^^).

game/ui-strings/src/main/kotlin/de/gleex/pltcmd/game/ui/strings/transformations/Unit.kt
Baret: Yup. It was quite hard to come up with all those short versions ^^

I thought there might be far more than 26 units so having a unique ICON-lengt string wouldn't be possible. But each `UnitKind` might fit into the alphabet so I came up with the restriction to make it unique per unit kind. The context or other distinctions in the UI would probably make it clear that this single letter represent an infantry unit for example.
I tried to incorporate a system for "classes" that appear in multiple unit kinds but I guess I didn't fully enforce it. There are fire support units or scouting units and I thought I would always represent the fire support unit with an "X" and the scouting unit with an "S" or something like that.

game/ui-strings/src/main/kotlin/de/gleex/pltcmd/game/ui/strings/transformations/Unit.kt
Baret: You got me! I wanted to use abbreviations like RTO, HMG and so forth exactly in those formats.

game/ui-strings/src/main/kotlin/de/gleex/pltcmd/game/ui/strings/transformations/Unit.kt
Baret: Thank you very much for that addition! As I explained in another comment I already tried to make it a bit systematic but that is a great improvement! Will do!

game/ui-strings/src/test/kotlin/de/gleex/pltcmd/game/ui/strings/transformations/UnitTransformationTest.kt
Baret: Nope. I test `Unit.toFrontendString` with `Units.toFrontendString` ;)

The blueprint should result in the same string as the actual unit (instance of the blueprint). They might each have an own transformation,  I don't care, but they should result in the same string (for now).

game/ui-strings/src/test/kotlin/de/gleex/pltcmd/game/ui/strings/transformations/UnitTransformationTest.kt
Baret: Hmmm, maybe. I thought the 1-char and 3-char lengths might be used as some kind of "ID"-style representation when the UI-designer is _really_ short on sparse space. But with 5-char strings you should be able to get it from the context. Or at least there it is easier to have unique values when writing the transformation.
I created the tests for ICON and SHORT3 because when extending `unitTransformation` it is quite hard for the dev to make out which strings are already taken. This way you simply run the test and see if your addition works. With SHORT5 it is rather easy to abbreviate the unit's name and most probably have it unique.

And to be honest I have a feeling that SHORT5 might not surviva all that long, maybe it simply is a silly format. But we will see that when we build more UI-elements.

Baret:
Thx for all that valuable feedback! I read all comments and answered/resolved where possible and will work on it when I get to it.

game/ui-strings/src/test/kotlin/de/gleex/pltcmd/game/ui/strings/transformations/DefaultTransformationTest.kt
Loomie: ah, ok.

Baret:
Ok now I'm done with my improvements. Have another look at it, please.

game/ui-strings/src/main/kotlin/de/gleex/pltcmd/game/ui/strings/transformations/Unit.kt
```diff
@@ -5,41 +5,56 @@ import de.gleex.pltcmd.game.ui.strings.Transformation
...
+/**
+ * Transformation for a unit blueprint [Units].
+ */
+internal val unitsTransformation: Transformation<Units> = { format ->
+    when(this) {
         Units.Rifleman            -> format.forUnit("Rifleman", "R", "RIF", "RifMa")
         Units.Grenadier           -> format.forUnit("Grenadier", "G", "GRE")
         Units.Officer             -> format.forUnit("Officer", "O", "OFF", "Off.")
-        Units.Radioman            -> format.forUnit("Radioman (RTO)", "P", "RAD")
+        Units.Radioman            -> format.forUnit("Radioman (RTO)", "C", "RTO", "RTO")
```
Loomie, line 28: OK, why is a "Radioman (RTO)" abbreviated to a "C" which is not contained anywhere? I think this type is elementary for the game: Give it the "R" from the Rifleman which in turn gets the "M" from the medic which can use  the "C". I find it more comprehensable, but I see that most time the first letter wins:

* C = radioman, _R_ifleman, _M_edic
* _R_adioman, rifle_M_an, medi_C_

Pretty hard with only one letter :/ So maybe just take the "R" from the rifleman and give him a free character like "F" (same like the "IVF"/"Infantry Fighting Vehicle"). Maybe it can also be a very simple character like a dot or dash!?

game/ui-strings/src/main/kotlin/de/gleex/pltcmd/game/ui/strings/transformations/Unit.kt
```diff
@@ -5,41 +5,56 @@ import de.gleex.pltcmd.game.ui.strings.Transformation
...
         Units.Medic               -> format.forUnit("Medic", "M", "MED")
         Units.CombatEngineer      -> format.forUnit("Combat engineer", "E", "ENG", "Engi.")
         Units.HMGTeam             -> format.forUnit("Heavy machine gun team (HMG)", "H", "HMG", "HMG")
-        Units.AntiTankTeam        -> format.forUnit("Anti-tank team", "T", "AT", "ATInf")
-        Units.AntiAirTeam         -> format.forUnit("Anti-air team", "A", "AA", "AAInf")
-        Units.Scout               -> format.forUnit("Infantry scout", "S", "SCI", "Scout")
+        Units.AntiTankTeam        -> format.forUnit("Anti-tank infantry", "T", "IAT", "ATInf")
+        Units.AntiAirTeam         -> format.forUnit("Anti-air infantry", "A", "IAA", "AAInf")
+        Units.Scout               -> format.forUnit("Scout infantry", "S", "ISC", "ScoIn")
```
Loomie, line 34: Why are 5 letters not "ScInf" to have the same suffix "Inf" as the ATs? Is "Sc" too short? Then maybe use "ScouI"? Maybe we need to see it in action - if five letters ever come into play.

Loomie (approved):
Looks good. The abbreviation of the units will never be easy to understand for everybody. We can polish them later.

game/ui-strings/src/main/kotlin/de/gleex/pltcmd/game/ui/strings/transformations/Unit.kt
```diff
@@ -0,0 +1,110 @@
...
+}
+
+/**
+ * Transformation for a unit blueprint [Units].
+ *
+ * There are some basic rules of thumb when abbreviating unit names, which may lead to counter intuitive 1-char
+ * or 3-char versions. They are only relevant for the formats [Format.ICON] and [Format.SHORT3]. [Format.SHORT5]
+ * should have enough space to create more or less readable short versions of unit names.
+ *
+ * - [Format.ICON] and [Format.SHORT3] abbreviations are always all-caps letters
```
Loomie, line 27:
```suggestion
 * - [Format.ICON] and [Format.SHORT3] abbreviations use capital letters only
```

Loomie (approved):

PR #117, opened 2020-10-04
Resolves #113 ranged combat

### LINKED ISSUE #113: Add range for combat
Currently two elements must be on adjacent tiles to fight. Add weapon ranges to enable fighting depending on weapons (like snipers can engage enemies from a larger distance).

Related:
#112

Loomie:


Loomie:
The calculations are done with range in mind. But currently only direct neighbors are attacked. So the effective fighting range is 100 m.

model/world/src/main/kotlin/de/gleex/pltcmd/model/world/coordinate/Coordinate.kt
```diff
@@ -58,6 +61,15 @@ data class Coordinate(val eastingFromLeft: Int, val northingFromBottom: Int) : C
         return above.eastingFromLeft == eastingFromLeft && northingFromBottom + 1 == above.northingFromBottom
     }
 
+    /** @return the euclidean distance */
+    infix fun distanceTo(other: Coordinate): Double {
```
Baret, line 65: Maybe with the unit classes in place, we could add an extension function which directly returns `Meters`. I think of an extension function because the model does not know how large a tile, represented by a coordinate, is, right? So it would need to reside somewhere near that definition.

Baret (commented):
Really impressive! All that math stuff :P
Great work, but a bit unintuitive to read and understand. What is a "precision"? What does "Precision.at100m" mean? I think that stuff just needs a bit better description by adding more documentation or renaming some stuff. Or at least use actual units like proposed in one of the review comments. All those doubles floating around... (pun intended).

model/combat/src/main/kotlin/de/gleex/pltcmd/model/combat/attack/Precision.kt
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * The area of a circle at the given range that is spread with this precision.
+     * @return area as squared unit of the given range's unit
+     */
+    fun areaAt(range: Double): Double = PI / 4.0 * offsetAt(range).pow(2.0)
+
+    /**
+     * @return the percentage to hit the given area at the given range with this precision
+     */
+    fun chanceToHitAreaAt(areaToHit: Double, range: Double): Double = (areaToHit / areaAt(range)).coerceAtMost(1.0)
```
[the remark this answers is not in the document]
Loomie: But the calculations are independent of the unit! It works the same for meters, inches or miles. ;) But both parameters need to have the same that's  true.

Models instead of primitive data types is just a good idea 👍

model/world/src/main/kotlin/de/gleex/pltcmd/model/world/coordinate/Coordinate.kt
Loomie: Yeah -> #120

PR #119, opened 2020-10-20
Resolves #118: text to speech

### LINKED ISSUE #118: Text to Speech
Have the log area read every line out aloud.

Possible Library MaryTTS: https://github.com/marytts/marytts

Baret:
I stumbled upon a text-to-speech library and thought about letting it read out the radio calls. And well, a little tinkering and... here we are :)

### DIFF
--- a/game/options/src/main/kotlin/de/gleex/pltcmd/game/options/GameOptions.kt
+++ b/game/options/src/main/kotlin/de/gleex/pltcmd/game/options/GameOptions.kt
@@ -60,4 +60,6 @@ object GameOptions {
      * Vertical number of sectors in the world.
      */
     const val SECTORS_COUNT_V: Int = 10
+
+    const val enableSound = false
 }
--- a/game/application/src/main/kotlin/de/gleex/pltcmd/game/application/examples/speaker/SpeakerExamples.kt
+++ b/game/application/src/main/kotlin/de/gleex/pltcmd/game/application/examples/speaker/SpeakerExamples.kt
@@ -1,0 +1,71 @@
+package de.gleex.pltcmd.game.application.examples.speaker
+
+import de.gleex.pltcmd.game.ui.sound.Speaker
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.runBlocking
+import marytts.server.Mary
+import org.hexworks.cobalt.logging.api.LoggerFactory
+
+private val log = LoggerFactory.getLogger("SpeakerExample")
+
+private val texts = listOf(
+        "Hello World!",
+        "Bravo, engage enemy at (1 1 5| 2 2 4), out."
+)
+
+private val effects = listOf(
+        "",
+//        "JetPilot",
+        "JetPilot+Rate(durScale:0.2)",
+        "JetPilot+Rate(durScale:0.5)",
+        "JetPilot+Rate(durScale:0.8)"
+)
+
+@ExperimentalCoroutinesApi
+fun main() {
+
+    /*
+    Documentation about maryTTS can be found here: http://marytts.phonetik.uni-muenchen.de:59125/documentation.html
+
+    Available effects: http://marytts.phonetik.uni-muenchen.de:59125/audioeffects
+    The list is:
+        Volume amount:2.0;
+        TractScaler amount:1.5;
+        F0Scale f0Scale:2.0;
+        F0Add f0Add:50.0;
+        Rate durScale:1.5;
+        Robot amount:100.0;
+        Whisper amount:100.0;
+        Stadium amount:100.0
+        Chorus delay1:466;amp1:0.54;delay2:600;amp2:-0.10;delay3:250;amp3:0.30
+        FIRFilter type:3;fc1:500.0;fc2:2000.0
+        JetPilot
+
+     Description for a specific effect (change effect name in the URL):
+        http://marytts.phonetik.uni-muenchen.de:59125/audioeffect-help?effect=TractScaler
+     */
+
+    log.info("Starting mary...")
+    Mary.startup()
+
+    for (effect in effects) {
+        Speaker.effects = effect
+        for (text in texts) {
+            log.info("Saying '$text' with effect $effect")
+            runBlocking {
+                Speaker.say(text)
+                delay(100)
+                Speaker.waitForQueueToEmpty()
+            }
+            log.info("")
+            log.info(" - - -")
+            log.info("")
+        }
+    }
+
+    runBlocking {
+        delay(500)
+        Speaker.waitForQueueToEmpty()
+    }
+}
--- a/game/application/src/main/resources/logback.xml
+++ b/game/application/src/main/resources/logback.xml
@@ -11,6 +11,7 @@
 <!--    <logger name="org.hexworks.amethyst.api.entity.Entity" level="DEBUG"/>-->
 
     <logger name="de.gleex.pltcmd.game.application" level="DEBUG"/>
+    <logger name="de.gleex.pltcmd.game.ui" level="DEBUG"/>
 
     <root level="INFO">
         <appender-ref ref="STDOUT"/>
--- a/game/ui/pom.xml
+++ b/game/ui/pom.xml
@@ -14,6 +14,10 @@
     <name>UI</name>
     <description>All the graphical stuff and user interaction.</description>
 
+    <properties>
+            <marytts.version>5.2</marytts.version>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>de.gleex.pltcmd.game</groupId>
@@ -38,8 +42,13 @@
 
         <dependency>
             <groupId>de.dfki.mary</groupId>
-            <artifactId>voice-cmu-slt-hsmm</artifactId>
-            <version>5.2</version>
+            <artifactId>voice-cmu-rms-hsmm</artifactId>
+            <version>${marytts.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>de.dfki.mary</groupId>
+            <artifactId>marytts-runtime</artifactId>
+            <version>${marytts.version}</version>
         </dependency>
     </dependencies>
 
--- a/game/ui/pom.xml
+++ b/game/ui/pom.xml
@@ -15,7 +15,7 @@
     <description>All the graphical stuff and user interaction.</description>
 
     <properties>
-            <marytts.version>5.2</marytts.version>
+            <marytts.version>5.2-beta3</marytts.version>
     </properties>
 
     <dependencies>
@@ -43,7 +43,7 @@
         <dependency>
             <groupId>de.dfki.mary</groupId>
             <artifactId>voice-cmu-rms-hsmm</artifactId>
-            <version>${marytts.version}</version>
+            <version>5.2</version>
         </dependency>
         <dependency>
             <groupId>de.dfki.mary</groupId>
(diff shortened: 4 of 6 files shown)


Baret:
Please have a review (before I can't stop myself and put even more in this "just trying smth real quick"-issue :D).
The different mary versions are required because in 5.2 the rate effect does not accelerate the voice output (aka the effect does not work).

Unfortunately currently the game is only startable from the IDE. When building with maven and starting the `application-0.2.0-SNAPSHOT-jar-with-dependencies.jar` you get a stacktrace:
```
Exception in thread "main" java.lang.ExceptionInInitializerError
        at de.gleex.pltcmd.game.application.Main.run(main.kt:58)
        at de.gleex.pltcmd.game.application.MainKt.main(main.kt:43)
        at de.gleex.pltcmd.game.application.MainKt.main(main.kt)
Caused by: marytts.exceptions.MaryConfigurationException: Cannot start MARY server
        at marytts.LocalMaryInterface.<init>(LocalMaryInterface.java:66)
        at de.gleex.pltcmd.game.ui.sound.speech.Speaker.<clinit>(Speaker.kt:79)
        ... 3 more
Caused by: java.lang.NullPointerException
        at marytts.util.MaryRuntimeUtils.ensureMaryStarted(MaryRuntimeUtils.java:70)
        at marytts.LocalMaryInterface.<init>(LocalMaryInterface.java:64)
        ... 4 more
```
:(

game/options/src/main/kotlin/de/gleex/pltcmd/game/options/GameOptions.kt
```diff
@@ -60,4 +60,6 @@ object GameOptions {
      * Vertical number of sectors in the world.
      */
     const val SECTORS_COUNT_V: Int = 10
+
+    const val enableSound = true
```
Loomie, line 64: Put it in upper case as the other constants.

game/sound/src/main/kotlin/de/gleex/pltcmd/game/sound/speech/Speaker.kt
```diff
@@ -0,0 +1,220 @@
...
+import org.hexworks.cobalt.logging.api.LoggerFactory
+import java.io.File
+import java.util.*
+import java.util.concurrent.atomic.AtomicBoolean
+import javax.sound.sampled.AudioSystem
+import javax.sound.sampled.DataLine
+import javax.sound.sampled.SourceDataLine
+
+/**
+ * Use this object for text-to-speech. All strings given to [say] wll be read aloud one after another. This means
```
Loomie, line 23:
```suggestion
 * Use this object for text-to-speech. All strings given to [say] will be read aloud one after another. This means
```

game/sound/src/main/kotlin/de/gleex/pltcmd/game/sound/speech/Speaker.kt
```diff
@@ -0,0 +1,220 @@
...
+            )
+
+            log.debug("Available voices: ${mary.availableVoices}")
+            val defaultVoice = Voice.getDefaultVoice(Locale.US)
+            log.debug("Default US voice: ${defaultVoice.name}")
+
+
+            playLoop = GlobalScope.launch {
+                log.debug("Launching play loop")
+                while (isActive) {
```
Loomie, line 96: This seems to be always true as "Stopped play loop" is never logged! I think an own scope is needed that can be cancelled.

game/sound/src/main/kotlin/de/gleex/pltcmd/game/sound/speech/Speaker.kt
```diff
@@ -0,0 +1,220 @@
...
+        if (GameOptions.SOUND_ENABLED.not()) {
+            return
+        }
+        if (IGNORED_PHRASES.any {
+                    text.contains(it, true)
+                }) {
+            log.debug("Ignored phrase detected, not saying '$text'")
+            return
+        }
+        val speechDirectory = File(FOLDER_SPEECH_FILES)
```
Loomie, line 157: `File` is soo before Java 7. `Path` and `Files` is the [new I/O 2](https://docs.oracle.com/javase/tutorial/essential/io/fileio.html).

But [`kotlin.io`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/) doesn't seem to know that!?

game/sound/src/main/kotlin/de/gleex/pltcmd/game/sound/speech/Speaker.kt
```diff
@@ -0,0 +1,220 @@
...
+        if (!speechDirectory.exists()) {
+            speechDirectory.mkdir()
+        }
+
+        // reuse existing texts if they had the same effects
+        mary.audioEffects = effects.toString()
+        val path = "${speechDirectory.absolutePath}/${text.hashCode()}_${mary.audioEffects.hashCode()}.wav"
```
Loomie, line 164: Please, use a model class (`File`/`Path`) instead of String! So you don't have to use the os specific path separator '/ '.

Loomie, line 164: The `hashCode()` method returns the same hash only in the same JRE runtime instance! It is not fix for multiple application startups!

game/sound/src/main/kotlin/de/gleex/pltcmd/game/sound/speech/Speaker.kt
```diff
@@ -0,0 +1,220 @@
...
+        if (!soundFile.exists()) {
+            log.debug("Creating new sound file $path")
+            val audio = mary.generateAudio(text)
+            log.debug("Generated speech audio with effects: ${mary.audioEffects}")
+            val samples = MaryAudioUtils.getSamplesAsDoubleArray(audio!!)
+            MaryAudioUtils.writeWavFile(samples, path, audio.format)
+        }
+        soundFile.deleteOnExit()
+
+        filenames.send(path)
```
Loomie, line 176: Use `soundFile` instead of `path` as String is not very useful.

game/sound/src/main/kotlin/de/gleex/pltcmd/game/sound/speech/Speaker.kt
```diff
@@ -0,0 +1,220 @@
...
+     * - JetPilot
+     */
+    internal var effects = EffectList.none
+
+    init {
+        if (GameOptions.SOUND_ENABLED) {
+            log.debug("Starting MaryTTS engine...")
+            mary = LocalMaryInterface()
+            if(Mary.currentState() == Mary.STATE_RUNNING) {
+                log.debug("MaryTTS started.")
```
Loomie, line 81: As you mention in the KDoc that the startup may take some time, it would be nice to log the time it took. For me currently the difference between the two log messages is about 1.500 ms.

game/application/src/main/kotlin/de/gleex/pltcmd/game/application/main.kt
```diff
@@ -50,6 +55,8 @@ open class Main {
     open fun run() {
         val application = SwingApplications.startApplication(UiOptions.buildAppConfig())
 
+        Speaker.startup()
```
Loomie, line 58: As this takes some time consider calling this asynchronous!

game/sound/src/main/kotlin/de/gleex/pltcmd/game/sound/speech/Speaker.kt
```diff
@@ -0,0 +1,220 @@
...
+        }
+        soundFile.deleteOnExit()
+
+        filenames.send(path)
+    }
+
+    private fun play(filename: String) {
+        val soundFile = File(filename)
+        if (soundFile.exists()) {
+            val audioStream = AudioSystem.getAudioInputStream(soundFile)
```
Loomie, line 182: We should check if reading files is faster than calculating the text-to-speech. Because disk i/o is relatively slow. But we could also cache the bytes in memory (for example with a fixed 10 MiB pool). But this are performance considerations.

game/sound/src/main/kotlin/de/gleex/pltcmd/game/sound/speech/Speaker.kt
```diff
@@ -0,0 +1,220 @@
...
+                    while (count != -1) {
+                        count = audioStream.read(buffer, 0, buffer.size)
+                        if (count >= 0) {
+                            it.write(buffer, 0, count)
+                        }
+                    }
+
+                    it.drain()
+                }
+            } finally {
```
Loomie, line 203: Looks like you want to lock the audio when in use. So instead of an AtomicBoolean I would suggest a corresponding mutual exclusion functionality from [java.util.concurrent](https://docs.oracle.com/javase/8/docs/api/index.html?java/util/concurrent/package-summary.html). Maybe a [`Semaphore`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Semaphore.html) is more approtiate as a [`Lock`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/Lock.html). The benefit is that you can just wait for it to be freed instead of the while-delay in `waitForQueueToEmpty()`.

Loomie: PS: maybe we can just use Jobs? They can be waited for (joined).

game/sound/src/main/kotlin/de/gleex/pltcmd/game/sound/speech/Speaker.kt
```diff
@@ -0,0 +1,220 @@
...
+    private fun play(filename: String) {
+        val soundFile = File(filename)
+        if (soundFile.exists()) {
+            val audioStream = AudioSystem.getAudioInputStream(soundFile)
+            val audioFormat = audioStream!!.format
+            val info = DataLine.Info(SourceDataLine::class.java, audioFormat)
+            val sourceDataLine = AudioSystem.getLine(info) as SourceDataLine
+            try {
+                playingSound.set(true)
+                sourceDataLine.use {
```
Loomie, line 188: This call or the lambda should be its own method. Something like `play(AudioInputStream, SourceDataLine)`.

game/sound/src/main/kotlin/de/gleex/pltcmd/game/sound/speech/Speaker.kt
```diff
@@ -0,0 +1,220 @@
...
+                    it.open(audioFormat)
+                    it.start()
+
+                    var count = 0
+                    val buffer = ByteArray(4096)
```
Loomie, line 193: Why 4096 instead of `sourceDataLine.getBufferSize()`?

game/sound/src/main/kotlin/de/gleex/pltcmd/game/sound/speech/Speaker.kt
Baret: They are deleted at (normal) JVM-termination anyways ;)

game/sound/src/main/kotlin/de/gleex/pltcmd/game/sound/speech/Speaker.kt
```diff
@@ -0,0 +1,220 @@
...
+ * advisable to call [startup] earlier as it might take some time and thus delay the output of the first text.
+ *
+ * How this object works in detail: [say] creates an audio file from the given text and saves it locally and then
+ * queues this file for replay. The file names are generated from the hashcode of the given text so the same
+ * text does not need to be written to a file again.
+ */
+@ExperimentalCoroutinesApi
+object Speaker {
+
+    private const val FOLDER_SPEECH_FILES = "./speech"
```
Loomie, line 36: remove the os specific "./" and just use the name.

Loomie (commented):
Looks and works good! For some implementation detail improvements see the comments.

PR #121, opened 2020-10-31
Resolves #81: Vision

### LINKED ISSUE #81: Implement vision
**Basic idea/big picture**
To be able to engage other elements they need to see each other. So every element (and probably other future entities, who knows) need field of vision (FOV).

**Details**
As a rather simple start every element should have the same field of vision range. It should be influenced by the surrounding terrain, especially terrain height. You just cannot look behind mountains. That means higher terrain blocks vision, when further terrain is lower.
It should suffice for the beginning when vision is "binary". An element can see a field of terrain or it can't. Maybe that is already enough and _detection_ of an enemy element present in that field is a second step affected by the enemy's state and surrounding terrain.

**Technical background**
Vision should be an attribute that can be given to any entity. The corresponding system checking the visible fields could simply "scan" all visible fields for other elements and trigger a report radio call (like "contact, unknown platoon at (x|y)/east of us").
Later this system will probably feed into the element's knowledge which other systems will use to decide what to do.

Maybe http://squidpony.github.io/SquidLib/squidlib-util/apidocs/squidpony/squidgrid/FOV.html can help with this. It should be investigated.

Loomie:


game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/behaviours/LookAround.kt
```diff
@@ -0,0 +1,76 @@
...
+        if (currentPosition != entity.lookingFrom || entity.visibleTiles.isEmpty) {
+            // position changed or view was reset
+            val visibleTiles = lookAround(currentPosition, context.world)
+            log.debug("Updating vision of ${(entity as ElementEntity).callsign} to $visibleTiles")
+            entity.updateVision(currentPosition, visibleTiles)
+        }
+        return true
+    }
+
+    private fun lookAround(from: Coordinate, world: WorldMap): CoordinateArea {
```
Baret, line 49: I could write a lenthy comment about what I think of this implementation and what might be done differently, but I guess it's easiest when we talk about it in discord real quick :)

Baret (changes requested):

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/behaviours/LookAround.kt
Loomie: Reworked with Commit 15e7f0cdfd7396c4809068b4a6582b3e033d04a2

model/world/src/main/kotlin/de/gleex/pltcmd/model/world/coordinate/CoordinateRectangle.kt
```diff
@@ -6,7 +6,7 @@ package de.gleex.pltcmd.model.world.coordinate
 class CoordinateRectangle(
         val bottomLeftCoordinate: Coordinate,
         val topRightCoordinate: Coordinate) :
-        CoordinateArea(bottomLeftCoordinate..topRightCoordinate) {
+        CoordinateArea({ (bottomLeftCoordinate..topRightCoordinate).toSortedSet() }) {
```
Baret, line 9: Don't we have the constructor with `Coordinate.Progression` for this?

model/world/src/main/kotlin/de/gleex/pltcmd/model/world/coordinate/MainCoordinate.kt
```diff
@@ -24,5 +24,17 @@ data class MainCoordinate(val eastingFromLeft: Int, val northingFromBottom: Int)
      */
     fun toCoordinate() = Coordinate(eastingFromLeft * TILE_COUNT, northingFromBottom * TILE_COUNT)
 
+    /**
+     * Sort from most south-west to most north-east. Going line wise first east and then north.
+     * Example: 2|2, 3|2, 1|3
+     */
+    override fun compareTo(other: MainCoordinate): Int {
```
Baret, line 31: This is currently copy&pasted from `Coordinate`. Can we create a generic comparator taking a northing and easting?

Baret (commented):
Just some small things. After that we can go to master.

Maybe I can fix those myself, but when you @Loomie get to it first, go ahead :)

model/world/src/main/kotlin/de/gleex/pltcmd/model/world/coordinate/CoordinateRectangle.kt
Loomie: But this constructor is lazy! I should have removed the other one...

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/renderers/SignalVisualizer.kt
```diff
@@ -0,0 +1,65 @@
...
+    private fun draw(signal: Signal<*>) {
+        currentJob = signal.all
+                .onEach { (coordinate, signalStrength) ->
+                    world.fetchBlockAt(coordinate)
+                            .ifPresent { block ->
+                                block.setOverlay(TileRepository.forSignal(signalStrength))
+                                currentBlocks.add(block)
+                            }
+                }
+                .launchIn(CoroutineScope(Dispatchers.Default))
```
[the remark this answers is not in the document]
Loomie: The default dispatcher uses different threads (confirmed with a `println`). I think the problem is the sequential nature of the flow. A buffer didn't help either.

Baret (approved):
Master, here we come!

PR #122, opened 2020-11-19
Trying to use improved signaling for vision

Baret:
Instead of writing a new line of sight algorithm to use it for vision I tried abstracting the radio signal code. In the end it is just a line of sight algo itself so why not reuse it? And we might later reuse it again for sound or maybe even more.


model/signals/core/src/main/kotlin/de/gleex/pltcmd/model/signals/core/SignalPower.kt
```diff
@@ -0,0 +1,9 @@
+package de.gleex.pltcmd.model.signals.core
+
+abstract class SignalPower(val power: Double) {
```
Loomie, line 3: Not a data class? I'm not sure if this should be abstract. The `initialProcessingValue` (which is a function and not a field?) can also be implemented directly in `Signal` where it is used. There the "processing" is specific. Also I think the function should be named something like `createFullStrength()` or `getMaxSignalPower()`. OK a radio may not transmit on full power, so maybe call it `emitSignal()` there to indicate that a new signal is created at the origin.

model/signals/core/src/main/kotlin/de/gleex/pltcmd/model/signals/core/SignalStrength.kt
```diff
@@ -0,0 +1,44 @@
+package de.gleex.pltcmd.model.signals.core
+
+/**
+ * Represents the remaining strength of a [Signal] at a specific location in percent. The [strength] must be
+ * a value between 0.0 and 1.0 inclusive. An exception is thrown if the given value is invalid.
+ */
+data class SignalStrength(
```
Loomie, line 7: Please use this new project also in `radio` to not duplicate all the classes. Currently the review is a bit hard to distinguish between moved, copied and new code.

model/signals/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+            <dependency>
+                <groupId>de.gleex.pltcmd.model.signals</groupId>
+                <artifactId>core</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>de.gleex.pltcmd.model.signals</groupId>
+                <artifactId>vision</artifactId>
+                <version>${project.version}</version>
+            </dependency>
```
Loomie, line 48: Add a separator comment like the on ein `model/pom.xml` to end the `modules` section:
```
<!-- other internal projects -->
```

model/signals/signals-all/pom.xml
```diff
@@ -0,0 +1,30 @@
...
+
+    <name>All signals dependencies</name>
+    <description>A simple aggregator module to get all signal modules with one dependency</description>
+
+    <packaging>pom</packaging>
+
+    <dependencies>
+        <dependency>
+            <groupId>de.gleex.pltcmd.model.signals</groupId>
+            <artifactId>core</artifactId>
```
Loomie, line 22: This dependency is also transient from `vision`. So I think this is not needed in the aggregator. Only the root projects of signal dependencies are required here.

model/signals/vision/src/main/kotlin/de/gleex/pltcmd/model/signals/vision/Vision.kt
```diff
@@ -0,0 +1,32 @@
...
+    override fun overAir(value: Double): Double {
+        return value - AIR_LOSS
+    }
+
+    override fun throughGround(value: Double): Double {
+        return VisionPower.MIN_POWER
+    }
+
+    override fun throughTerrain(value: Double, terrainType: TerrainType): Double {
+        return value - 0.3
```
Loomie, line 26: Is this final or is a `TODO` missing that depends on the terrain type?

model/signals/vision/src/main/kotlin/de/gleex/pltcmd/model/signals/vision/Vision.kt
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    override fun toSignalStrength(calculationValue: Double): SignalStrength =
+            SignalStrength(max(calculationValue, SignalStrength.MIN_ALLOWED_VALUE))
```
Loomie, line 30: Why is this not directly in the `SignalStrength` constructor? The field for the minimum value is also contained there. So this function wants to be in SignalStrength where it is just the constructor. Then this function can also be removed from `PropagationModel`.

model/signals/vision/src/main/kotlin/de/gleex/pltcmd/model/signals/vision/VisionPower.kt
```diff
@@ -0,0 +1,24 @@
...
+
+    companion object {
+        val MIN_POWER = 0.0
+        val MIN = VisionPower(MIN_POWER)
+    }
+
+    operator fun minus(i: Double): VisionPower =
+            VisionPower(power - i)
+
+    override fun initialProcessingValue(): Double = 1.0
```
Loomie, line 19: The KDoc says "range in tiles" but it is a value between 0.0 and 1.0? So where it is mapped to more than one tile? As `SignalPower` just wraps a `Double` I don't know if we need special sub types. The minium and maxium allowed values are provided by the PropagationModel anyways.

model/signals/vision/src/main/kotlin/de/gleex/pltcmd/model/signals/vision/VisualSignal.kt
```diff
@@ -0,0 +1,19 @@
...
+import de.gleex.pltcmd.model.signals.core.Signal
+import de.gleex.pltcmd.model.world.WorldArea
+import de.gleex.pltcmd.model.world.coordinate.Coordinate
+import kotlin.math.roundToInt
+
+class VisualSignal(
+        override val origin: Coordinate,
+        area: WorldArea,
+        power: VisionPower,
+        override val model: Vision
```
Loomie, line 12: OK, I asked myself when viewing at `Signal` already: Why are two values abstract and two are fields of `Signal`. Now I see that there is no difference at all. I think all fields should be kept in the super class where they are defined already.

Also: Why is `Signal` abstract? This class adds no functionality. Just provide the values to the constructor.

model/world/src/main/kotlin/de/gleex/pltcmd/model/world/Sector.kt
```diff
@@ -8,7 +8,7 @@ import kotlin.random.Random
 /**
  * A sector has 50 by 50 [WorldTile]s (it is a square).
  */
-data class Sector(val origin: Coordinate, val tiles: SortedSet<WorldTile>) : Comparable<Sector> {
+class Sector(val origin: Coordinate, tiles: SortedSet<WorldTile>) : Comparable<Sector>, WorldArea(tiles) {
```
Loomie, line 11: When removing `data` add all the generated functions like `equals/hashCode`!

model/world/src/main/kotlin/de/gleex/pltcmd/model/world/WorldArea.kt
```diff
@@ -0,0 +1,41 @@
+package de.gleex.pltcmd.model.world
+
+import de.gleex.pltcmd.model.world.coordinate.Coordinate
+import de.gleex.pltcmd.model.world.coordinate.CoordinateArea
+import de.gleex.pltcmd.model.world.coordinate.CoordinatePath
+import org.hexworks.cobalt.datatypes.Maybe
+import java.util.*
+
+/**
+ * A part of the [WorldMap] containing a set of [WorldTile]s. As world tiles map a terrain to a coordiante, a world
```
Loomie, line 10:
```suggestion
 * A part of the [WorldMap] containing a set of [WorldTile]s. As world tiles map a terrain to a coordinate, a world
```

model/world/src/main/kotlin/de/gleex/pltcmd/model/world/WorldMap.kt
```diff
@@ -82,6 +83,17 @@ data class WorldMap private constructor(private val originToSector: SortedMap<Co
...
 
+    /**
+     * Returns a [WorldArea] containing all [WorldTile]s contained in the given [CoordinateArea]
+     */
+    fun areaOf(coordinateArea: CoordinateArea): WorldArea {
+        return WorldArea(sectors
+                .filter { sector ->  sector.origin in coordinateArea.sectorOrigins }
+                .flatMap { sector -> sector.tiles }
+                .filter { worldTile -> worldTile.coordinate in coordinateArea }
+                .toSortedSet())
```
Loomie, line 94: Maybe the other way around is more performant; Instead of filtering the whole world just map the area to its WorldTiles. The sector.origin optimization may be enough or even better to optimize the performance. We will see...

model/world/src/main/kotlin/de/gleex/pltcmd/model/world/coordinate/CoordinateArea.kt
```diff
@@ -14,10 +15,35 @@ open class CoordinateArea(private val coordinates: SortedSet<Coordinate>) : Iter
     val isEmpty: Boolean
         get() = coordinates.isEmpty()
 
+    /**
+     * Maps all [MainCoordinate]s contained in this area to their [Coordinate]s that are present in this
+     * area. This map is useful to more efficiently find coordinates in very large areas.
+     */
+    protected val mainCoordinatesMap: Map<MainCoordinate, SortedSet<Coordinate>> =
```
Loomie, line 22: The values are never used (currently). Also `mainCoordinates` and `sectorOrigins` do the same. I think sector origins should be used as they are more fine grained and so the purpose of filtering is more effective.

Loomie (changes requested):
Not bad, but needs some polishing.

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/entities/EntitySet.kt
```diff
@@ -0,0 +1,66 @@
...
+
+    fun allElements(): EntitySet<ElementType> = ofType(ElementType::class)
+
+    fun inArea(area: CoordinateArea): EntitySet<Positionable> =
+            ofType(Positionable::class)
+                    .filter { it.currentPosition in area }
+                    .toEntitySet()
+
+    fun without(vararg entities: GameEntity<T>): EntitySet<T> =
+            allEntities.removeAll(entities.toList())
```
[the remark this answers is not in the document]
Baret: It is a `PersistentSet`, all modifications return a new set. So here a new set is being created and then "wrapped up" into an `EntitySet`.

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/entities/EntitySet.kt
```diff
@@ -0,0 +1,66 @@
...
+
+    override fun contains(element: GameEntity<T>): Boolean =
+            allEntities.contains(element)
+
+    override fun containsAll(elements: Collection<GameEntity<T>>): Boolean =
+            allEntities.containsAll(elements)
+
+    override fun isEmpty(): Boolean =
+            allEntities.isEmpty()
+}
```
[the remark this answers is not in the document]
Baret: You might be right. I got closer and closer to "this is simply a set" while coding it ^^

game/pom.xml
```diff
@@ -102,6 +102,12 @@
                 <artifactId>world</artifactId>
                 <version>${project.version}</version>
             </dependency>
+            <dependency>
+                <groupId>de.gleex.pltcmd.model.signals</groupId>
+                <artifactId>signals-all</artifactId>
+                <type>pom</type>
```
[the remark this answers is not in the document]
Baret: It does not contain any code, it is simply a shortcut for "core, vision, radio and possibly more", so that you only need one dependency instead of 2 or 3 or later even more. I was not sure how to create such an "aggregator module" (it is the wrong term in maven speak).

model/signals/core/src/main/kotlin/de/gleex/pltcmd/model/signals/core/PropagationModel.kt
```diff
@@ -0,0 +1,16 @@
+package de.gleex.pltcmd.model.signals.core
+
+import de.gleex.pltcmd.model.world.terrain.TerrainType
+
+interface PropagationModel<P: SignalPower> {
```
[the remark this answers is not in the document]
Baret: I had a conceptual problem here. I wanted to have the initial `SignalPower` in the model so it can be used as reference in `throughTerrain´ and other calls. It would be turned into a `SignalStrength` immediately and the method calls would return a reduced `SignalStrength`. But somehow that didn't work out for how I wanted to implement vision so I made it like this:

In the beginning there is a `SignalPower`, then it is turned into some arbitrary `Double` to calculat the signal loss along the way, and then this double is turned into a remaining `SignalStrength`. But I gotta think about that again...

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/facets/Detects.kt
```diff
@@ -0,0 +1,40 @@
...
+import org.hexworks.amethyst.api.Command
+import org.hexworks.amethyst.api.Consumed
+import org.hexworks.amethyst.api.Pass
+import org.hexworks.amethyst.api.Response
+import org.hexworks.amethyst.api.base.BaseFacet
+import org.hexworks.amethyst.api.entity.EntityType
+import org.hexworks.amethyst.api.extensions.responseWhenCommandIs
+import org.hexworks.cobalt.logging.api.LoggerFactory
+
+object Detects: BaseFacet<GameContext>(
```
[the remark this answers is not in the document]
Baret: Coming soon (tm) ;P

model/signals/vision/src/main/kotlin/de/gleex/pltcmd/model/signals/vision/Vision.kt
Baret: Nope, that was just some quick "Gotta put some numbers here". Details to follow.

model/world/src/main/kotlin/de/gleex/pltcmd/model/world/WorldMap.kt
Baret: I want to write a little performance test and run it with those optimizations and without (on the master branch).

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/GameContext.kt
```diff
@@ -17,12 +19,12 @@ import kotlin.random.Random
...
         val currentTick: TickId,
         val world: WorldMap,
-        private val allElements: Set<ElementEntity>,
+        val entities: EntitySet<EntityType>,
         val random: Random
 ) : Context {
 
     fun findElementAt(location: Coordinate): Maybe<ElementEntity> {
-        return Maybe.ofNullable(allElements.find { it.currentPosition == location })
+        return Maybe.ofNullable(entities.allElements().find { it.currentPosition == location })
```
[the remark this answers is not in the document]
Baret: We should simply move both methods to `EntitySet`. Which increases its value ;)

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/entities/EntitySet.kt
```diff
@@ -0,0 +1,66 @@
...
+import de.gleex.pltcmd.game.engine.extensions.GameEntity
+import de.gleex.pltcmd.model.world.coordinate.CoordinateArea
+import kotlinx.collections.immutable.PersistentSet
+import kotlinx.collections.immutable.toPersistentSet
+import org.hexworks.amethyst.api.entity.EntityType
+import kotlin.reflect.KClass
+
+class EntitySet<T : EntityType>(initialSet: Set<GameEntity<T>> = emptySet()) : Set<GameEntity<T>> {
+
+    private var allEntities: PersistentSet<GameEntity<T>> = initialSet.toPersistentSet()
```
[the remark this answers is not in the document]
Baret: I can not reassign the `PersistentSet` when I use it a delegate. But as it is immutable (mutating operations return a new instance) it needs to be a `var`. 
And yes, it needs to be a `PersistentSet` as I already ran into a `ConcurrentModificationException` when playing around with it (it was a simple `Set` before).

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/entities/EntitySet.kt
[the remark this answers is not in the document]
Baret: After some improvement and additions I think we should keep it. There would be _so_ many extension functions that it makes sense for me to couple them together in an own class. As it extends `Set` you can use the default methods on it anyways.
And I would like to add performance logging later which is also best when we have some kind of `performanceLogger` field in the class.

game/engine/src/test/kotlin/de/gleex/pltcmd/game/engine/systems/behaviours/LookAroundTest.kt
```diff
@@ -30,7 +28,7 @@ class LookAroundTest : StringSpec({
         }
     }
     val sector = Sector(origin, tiles)
-
+/*
```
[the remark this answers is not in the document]
Baret: Yup, Unittests are still a TODO.

model/signals/core/src/main/kotlin/de/gleex/pltcmd/model/signals/core/Signal.kt
```diff
@@ -0,0 +1,57 @@
...
+            }
+        }
+        return model.toSignalStrength(currentPower)
+    }
+
+    // evtl hat das noch ein voll imperformantes
+//    fun signalMap(): Map<Coordinate: Double> {
+//        // das eine Linie zu jedem Punkt in der Reichweite zieht
+//        // (so wie das jetzt der SignalVisualizerFragment oder wie das Dingen heisst macht)
+//    }
```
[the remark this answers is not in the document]
Baret: Nope, use it ;)

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/entities/EntitySet.kt
```diff
@@ -4,43 +4,92 @@ import de.gleex.pltcmd.game.engine.entities.types.ElementType
...
+    fun elementsAt(location: Coordinate): EntitySet<ElementType> =
+            allElements()
+                    .filter { it.currentPosition == location }
+                    .toEntitySet()
 
+    /**
+     * @return a [Maybe] containing the first [GameEntity] of type [ElementType] found at the given [location]
+     *  or an empty [Maybe] if no elements are present.
+     */
+    fun firstElementAt(location: Coordinate): Maybe<GameEntity<ElementType>> =
```
Loomie, line 60: We have a typealias for `GameEntity<ElementType>` it is named `ElementEntity` ;)

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/entities/types/Seeing.kt
```diff
@@ -16,17 +18,15 @@ typealias SeeingEntity = GameEntity<Seeing>
 private val SeeingEntity.visionAttribute: VisionAttribute
     get() = getAttribute(VisionAttribute::class)
 
-internal var SeeingEntity.vision: VisualSignal
+// visible for the UI
+var SeeingEntity.vision: Vision
```
Loomie, line 22: As it is a `var` it is not only visble but also editable!

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/renderers/SignalVisualizer.kt
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    private fun draw(signalMap: Map<Coordinate, SignalStrength>) {
+        signalMap.forEach { (coordinate, signalStrength) ->
+            world.fetchBlockAt(coordinate)
+                    .ifPresent { block ->
+                        block.setOverlay(TileRepository.forSignal(signalStrength))
+                    }
+        }
+        lastDrawnMap = signalMap
```
Loomie, line 46: Why not directly remember the blocks? The values of the map are never used.

model/communication/pom.xml
```diff
@@ -7,11 +7,10 @@
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
-    <artifactId>radio</artifactId>
+    <artifactId>communication</artifactId>
```
Loomie, line 10: The packes have not been changed. It is not consistent with the other projects where the package reassembles the project hierarchy:
* Folder: <pltcmd>/model/communication
* Package: ...pltcmd.model.radio

Maybe we can switch the package names of `communication` and `radio`:
`de.gleex.pltcmd.model.radio.communication`
gets
`de.gleex.pltcmd.model.communication.radio`

model/signals/core/src/main/kotlin/de/gleex/pltcmd/model/signals/core/SignalPropagator.kt
```diff
@@ -0,0 +1,41 @@
...
+     */
+    fun signalLossThroughAir()
+
+    /**
+     * Called for each tile a signal travels through ground. This usually accumulates the most signal loss.
+     */
+    fun signalLossThroughGround()
+
+    /**
+     * Called for each tile a signals travels through the given [TerrainType].
```
Loomie, line 38:
```suggestion
     * Called for each tile a signal travels through the given [TerrainType].
```

model/signals/radio/src/main/kotlin/de/gleex/pltcmd/model/signals/radio/RadioSignalPropagator.kt
```diff
@@ -0,0 +1,86 @@
...
+ * - When the remaining power reaches [MIN_POWER_THRESHOLD] the signal is cut off and treated as [SignalStrength.NONE]
+ */
+class RadioSignalPropagator(initialRadioPower: Double) : SignalPropagator {
+
+    companion object {
+        /**
+         * The maximum number of tiles a radio signal can travel with the given power. This is only possible
+         * when it only travels through air.
+         */
+        fun maxRangeInTiles(power: Double): Int {
```
Loomie, line 26: Put the function below the fields/constants.

model/signals/vision/src/main/kotlin/de/gleex/pltcmd/model/signals/vision/VisionPropagator.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    override val remainingSignalStrength: SignalStrength
+        get() = (1.0 - accumulatedSignalLoss).toSignalStrength()
+
+    override val signalDepleted: Boolean
+        get() = remainingTiles <= 0 || accumulatedSignalLoss >= 1.0
+
+    override fun signalLossThroughAir() {
+        if (log.isTraceEnabled()) {
+            log.trace("\t\tSignal travels through air")
```
Loomie, line 35:
```suggestion
            log.trace("\t\Vision travels through air")
```

model/signals/vision/src/main/kotlin/de/gleex/pltcmd/model/signals/vision/builder/WorldMapExtension.kt
```diff
@@ -1,15 +1,26 @@
...
+ * Creates a new [Signal] representing the vision of an entity at [location] with the given [visualRange].
+ *
+ * @param location from where someone is viewing
+ * @param visualRange the maximum range that can possibly be observed
+ *
+ * @return a visual
+ */
+fun WorldMap.visionAt(location: Coordinate, visualRange: VisionPower): Vision {
+    val radius: Int = ceil(visualRange.power).toInt()
+    return Signal(
```
Loomie, line 21:
```suggestion
    return Vision(
```

model/signals/vision/src/main/kotlin/de/gleex/pltcmd/model/signals/vision/builder/WorldMapExtension.kt
```diff
@@ -1,15 +1,26 @@
...
+            location,
+            areaOf(CoordinateCircle(location, radius)),
```
Loomie, line 23: I think we need a `circleAt()` besides `areaOf()` in `WorldMap`.

Loomie: Done in 63f252f

model/communication/pom.xml
Baret: There will be no `radio` package in the `communication` module as it is being moved over to the new `signals.radio` module. But I only started with that.

If you like go ahead and delete the old package and fix up the compile errors :D

Baret:
Thanks for the review! I will now go on to move the radio stuff over to the new signals module and delete the old stuff. After that I gotta move/fix/write some tests and then this PR is actually ready to be reviewed and merged back into your branch ;)
This just started as "little experiment"... oh well...

model/communication/pom.xml
Baret: Done.

model/communication/src/test/kotlin/de/gleex/pltcmd/model/radio/RadioSenderTest.kt
```diff
@@ -1,18 +1,6 @@
...
-import de.gleex.pltcmd.model.world.coordinate.CoordinateRectangle
-import de.gleex.pltcmd.model.world.terrain.Terrain
-import de.gleex.pltcmd.model.world.terrain.TerrainHeight
-import de.gleex.pltcmd.model.world.terrain.TerrainType
-import de.gleex.pltcmd.model.world.testhelpers.sectorAtWithTerrain
-import io.kotest.core.spec.style.StringSpec
-import io.kotest.matchers.shouldBe
-import org.hexworks.cobalt.databinding.api.extension.toProperty
-
+/*
```
Loomie, line 3: TODO implement or move tests

model/communication/src/test/kotlin/de/gleex/pltcmd/model/radio/broadcasting/RadioSignalTest.kt
```diff
@@ -0,0 +1,68 @@
+package de.gleex.pltcmd.model.radio.broadcasting
+
+import de.gleex.pltcmd.model.world.terrain.Terrain
+import de.gleex.pltcmd.model.world.terrain.TerrainHeight
+import de.gleex.pltcmd.model.world.terrain.TerrainType
+
+//class RadioSignalTest: WordSpec() {
```
Loomie, line 7: TODO update tests

model/communication/src/test/kotlin/de/gleex/pltcmd/model/radio/broadcasting/SignalStrengthTest.kt
```diff
@@ -0,0 +1,55 @@
+package de.gleex.pltcmd.model.radio.broadcasting
+
+//class SignalStrengthTest : WordSpec({
```
Loomie: TODO update tests

model/communication/src/test/kotlin/de/gleex/pltcmd/model/radio/testhelpers/Matchers.kt
```diff
@@ -0,0 +1,7 @@
+package de.gleex.pltcmd.model.radio.testhelpers
+
+// - - - Matchers for SignalStrength
+
+//infix fun SignalStrength.shouldBeExactly(expected: Double) = this.strength shouldBeExactly expected
```
Loomie, line 5: Remove?

model/world/src/main/kotlin/de/gleex/pltcmd/model/world/Sector.kt
Loomie: This is not resolved!

model/world/src/main/kotlin/de/gleex/pltcmd/model/world/coordinate/CoordinateArea.kt
Loomie: >  This conversation was marked as resolved by Baret 

Why? -> Reopen.

Loomie (changes requested):
Found "something" :D

model/world/src/main/kotlin/de/gleex/pltcmd/model/world/WorldMap.kt
Baret: Opened #127 for this.

model/world/src/main/kotlin/de/gleex/pltcmd/model/world/Sector.kt
Baret: Oops :)
done in 724bd81

model/world/src/main/kotlin/de/gleex/pltcmd/model/world/coordinate/CoordinateArea.kt
Baret: I think I wanted to provide future subclasses a way of more efficiently get coordinates. But yes, YAGNI.

Done in 55beba3

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/Game.kt
```diff
@@ -39,29 +45,34 @@ data class Game(val engine: Engine<GameContext>, val world: WorldMap, val random
...
     }
 
-    private fun removeElement(element: ElementEntity) {
-        log.debug("Removing element ${element.callsign} from game")
-        allElements.remove(element)
-        engine.removeEntity(element)
+    @Suppress("UNCHECKED_CAST")
+    private fun removeEntity(entity: AnyGameEntity) {
+        val entityName: String = if(entity.type == ElementType) {
+            (entity as ElementEntity).callsign.toString()
```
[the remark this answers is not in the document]
Baret: ElemenEntity has a `name` field? Or what do you mean?

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/Game.kt
```diff
@@ -74,7 +85,7 @@ data class Game(val engine: Engine<GameContext>, val world: WorldMap, val random
             affiliation: Affiliation = Affiliation.Unknown
     ): ElementEntity {
         val elementPosition = positionInSector.toProperty()
-        val radioSender = RadioSender(elementPosition, GameOptions.defaultRadioPower, world)
+        val radioSender = RadioSender(elementPosition, RadioPower(2000.0), world)
```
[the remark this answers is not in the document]
Baret: The idea of presets is great! I added some in 90e3aba but they are not yet in use.

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/Game.kt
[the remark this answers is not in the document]
Baret: Ah, the interface `Entity` has a field `name` documented with:
> Shorthand for accessing EntityType.name.

But as we only use typealiases for the entitytypes, we can not directly override `val name`. So in this case it makes more sense to get the callsign for a meaningful log message than to simply use `entity.name` as it would simply return "element" for every `ElementEntity`.

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/entities/EntitySet.kt
```diff
@@ -0,0 +1,119 @@
...
+            ofType<Positionable>()
+                    .filter { it.currentPosition in area }
+                    .toEntitySet()
+
+    /**
+     * @param entities the entities to exclude
+     *
+     * @return a new [EntitySet] containing all the entitie of this one except [entities].
+     */
+    fun without(vararg entities: GameEntity<T>): EntitySet<T> =
```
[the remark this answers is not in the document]
Baret: I tried that but stumbled upon a typing problem. To have the code like your snippet you would need to define `without` outside of the class. So it is independent of an actual entity set. But then you do not know the type of entities which means you would need to define it as
```kotlin
fun without(vararg entities: AnyGameEntity): (AnyGameEntity) -> Boolean {
// ...
}
```

This means you could provide entites of any type which may have no effect at all. While in the current implementation you can only filter out entites that have the same type as the ones contained in the set.
And I like the chaining style it currently has, although I see your point that we create new collections over and over again while doing so.

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/entities/EntitySet.kt
```diff
@@ -0,0 +1,119 @@
...
+                    .filter { it.currentPosition == location }
+                    .toEntitySet()
+
+    /**
+     * @return a [Maybe] containing the first [GameEntity] of type [ElementType] found at the given [location]
+     *  or an empty [Maybe] if no elements are present.
+     */
+    fun firstElementAt(location: Coordinate): Maybe<ElementEntity> =
+            // This should be faster than using elementsAt.first() because not the whole list needs to be filtered
+            Maybe.ofNullable(allElements().firstOrNull { it.currentPosition == location })
```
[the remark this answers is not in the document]
Baret: Great! 👍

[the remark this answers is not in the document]
Baret: And as I found out, `Sequence` is what you want to use when laziliy filtering large collections. And as the "root entity set" in `Game` contains _all_ entities I guess it will become pretty large at some point.

game/ui/src/main/kotlin/de/gleex/pltcmd/game/ui/fragments/tileinformation/SignalVisualizationFragment.kt
```diff
@@ -0,0 +1,67 @@
...
+
+    private val values: List<Pair<String, (SeeingEntity) -> Unit>> = listOf(
+            "None" to { visualizer.deactivate() },
+            "Vision" to { entity: SeeingEntity ->
+                visualizer.activateWith(entity.visionImmutable)
+            },
+            "Radio" to {
+                @Suppress("UNCHECKED_CAST")
+                if (it.type is Communicating) {
+                    it as CommunicatingEntity
```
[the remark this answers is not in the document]
Baret: Because it is not nessecary in kotlin :P

Baret:
I hope I got all your comments fixed. Maybe something got lost on the way because I did not resolve your comments as I guess that is your job now in the re-review.

I kept the tests out because most of this stuff is only well testable with #129. 

It is astonishing how much simpler the `RadioSender` and `BroadcastEvent` stuff became now that basically everything except the `Transmission` is contained in the signal :)

Issue #62, opened 2020-05-02
Implement Faction


Baret:
**Basic idea/big picture**
The player (and in the future other players or AI enemies) controls a `Faction`. It contains everything that belongs to one army. At least:
- An HQ with a fixed position
- All elements
- much more that may be left out for now
  - For example (just to have it written down already)
  - Fog of War or rather "Known terrain". Needed by the map view to know what to show at each tile
  - Relations to other factions
  - Knowledge of the faction that is inherited by elements that leave the base, see #104 

**Details**
I am not sure if `Faction` will be an `Entity` of its own or just an `Attribute` of all entities affiliated with a faction. Maybe the overarching entity will be the `Army` having attributes like `faction`, `relations` (to other factions), `elements`, `hq` and so on...

Loomie:
The base for the use of factions is done. It can be set as `Attribute` for an entity.

>  It contains everything that belongs to one army.

That has to be added in future issues. Maybe in the mentioned form:

> Maybe the overarching entity will be the `Army` having attributes like `faction`, `relations` (to other factions), `elements`, `hq` and so on...

Issue #104, opened 2020-08-14
Knowledge is power!


Baret:
**Basic idea/big picture**
Factions (#62 ) and elements should have "knowledge". It contains everything they know (well, duh). That means all their actions are based on it and when their knowledge is outdated or incomplete they may do things not in the ideal way or are unable altogether.

## REST OF ISSUE DESCRIPTION INCOMPLETE [TBD]

**Details**
What are the details? Is there anything that must not be missing?

**Technical background**
Do already know any technical details, for example where to put it or how to implement the feature?

**Related issues/Prerequisites**
The knowledge is also a part of a faction, which comes with #62 but maybe this can also be implemented before that issue is done

Loomie:
I try to create the interface use case driven almost like test-driven. I would like to access the knowledge like this:
```kotlin
val knownTerrain = entity.knowledgeAbout(MapTerrain)
```
If the knowledge model is mutable it can be modified with the same access. In the terrain example it could be `knownTerrain[coordinate] = seenTerrain`.

For the management of the knowledge we need maybe something like:
```kotlin
entity.remember[MapTerrain] = knownTerrain

entity.forget[MapTerrain]
```
For the location of units which change often and individually it might be:
```kotlin
entity.knowledgeAbout[Elements].update(elementInfo)
```

Hm, but for the time component the context or tick must be provided. So the knowledge can store the date of the knowledge which can be used to clean up obsolete data. Maybe:
```kotlin
entity.remember[MapTerrain, context.currentTick] = knownTerrain
entity.forgetObsolete(Elements, context.currentTick)
```
But if it can be forgotten the return value should be a `Maybe` when accessing knowledge. But that disturbs the clean accessor. So the KnowledgeType should provide a default value for "no knowledge". The type must also provide the maximum age before data becomes obsolete.
```kotlin
interface KnowledgeType<T> {
    val maxAgeInSeconds: Long
    fun emptyKnowledge(): T
}
```

That is only for `Knowledge` itself. The knowledge transfer is missing in my use cases. If the knowledge itself is just a domain object we need to define how to merge it. I think that would also be part of `KnowledgeType` that gets two objects and returns the common knowledge:
```kotlin
    fun merge(base: T, addition: T): T
```
That could be called in an extension function of `T` so we have the same as @Baret wrote above: `fun T.mergeWith(otherKnowledge: T)`.

Loomie:
Putting my what and Barets how together:

[...]

PR #138, opened 2021-02-07
Issue #62 Faction

Loomie:
Affiliation is now based on the faction.

### DIFF
--- a/model/faction/src/main/kotlin/de/gleex/pltcmd/model/faction/Faction.kt
+++ b/model/faction/src/main/kotlin/de/gleex/pltcmd/model/faction/Faction.kt
@@ -1,6 +1,4 @@
-package de.gleex.pltcmd.model.faction.de.gleex.pltcmd.model.faction
-
-import de.gleex.pltcmd.model.faction.FactionRelations
+package de.gleex.pltcmd.model.faction
 
 /**
  * One party in a military conflict.
@@ -8,3 +6,5 @@ import de.gleex.pltcmd.model.faction.FactionRelations
 data class Faction(val name: String) {
     val relations: FactionRelations = FactionRelations(this)
 }
+
+val UNIDENTIFIED = Faction("unidentified")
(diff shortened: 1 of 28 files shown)


Loomie:
I created a simple model that can be used in entities. Currently there is an attribute that holds a faction. That is already used in ElementEntity, CommunicationEntity and FOBEntity.

model/faction/src/main/kotlin/de/gleex/pltcmd/model/faction/Faction.kt
```diff
@@ -0,0 +1,8 @@
+package de.gleex.pltcmd.model.faction
+
+/**
+ * One party in a military conflict.
+ */
+data class Faction(val name: String)
+
+val UNIDENTIFIED = Faction("unidentified")
```
Baret, line 8: As we have already talked about it: This faction is not necessary. Factions and their relations to each other are facts from the beginning of the game. There is a fixed number of factions when a game starts and each faction needs to have a relation (allied, neutral, hostile) to every other faction. Btw. **this should be verified on game start**, or can we build a `Faction` constructor that implicitly checks that? Factions would need to be singletons then.

However, when the faction of an element is "unidentified", that is the _knowledge_ of a single entity. Another entity might well be aware which faction the element belongs to. So what you tried to achieve here with `UNIDENTIFIED` comes with #104 and for simplicity's sake we should simply leave it out for now. Else we would need to remove it later on which might cause confusion or hacky code.

Baret (commented):
It does not quite hit the desciption of #62 but I like the simplicity. It contains the core model for what a faction is and we can build up on this. This means we can forther down the road use it for gameplay features like the high command, friend/foe detection and so on.

But I'd like to have it without the "unidentified" faction as it feels like a hack we are going to solve with future tickets. And maybe it would be good to validate the faction relations on game start. Because what happens if you try to identify another element, you see it belongs to "Faction A" but you can not look up the relation of your own faction to that one? Shoot? Ignore them? ;)

Loomie:
> And maybe it would be good to validate the faction relations on game start

That is for the user of the faction that creates multiple instances. I don't see how the model can do this as it does not about _all_ factions.

> or can we build a `Faction` constructor that implicitly checks that?

I don't want to mix FactionRelations into Faction - no dependency cycles! Also I think we can't do such a hard check as one Faction after the other is created and not all at once.

PR #139, opened 2021-03-06
Issue 104: knowledge of contacts

Loomie:
Model for contacts. Contacts are remembered in an Attribute. Partial matching for more/less detailed contacts.

### DIFF
--- a/game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/facets/Detects.kt
+++ b/game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/facets/Detects.kt
@@ -3,17 +3,24 @@ package de.gleex.pltcmd.game.engine.systems.facets
 import de.gleex.pltcmd.game.engine.GameContext
 import de.gleex.pltcmd.game.engine.attributes.PositionAttribute
 import de.gleex.pltcmd.game.engine.attributes.VisionAttribute
+import de.gleex.pltcmd.game.engine.attributes.knowledge.LocatedContact
 import de.gleex.pltcmd.game.engine.entities.types.*
 import de.gleex.pltcmd.game.engine.extensions.logIdentifier
 import de.gleex.pltcmd.game.engine.messages.DetectEntities
 import de.gleex.pltcmd.game.engine.messages.DetectedEntity
+import de.gleex.pltcmd.model.elements.Contact
+import de.gleex.pltcmd.model.elements.Corps
+import de.gleex.pltcmd.model.elements.ElementKind
+import de.gleex.pltcmd.model.elements.Rung
 import de.gleex.pltcmd.model.signals.core.SignalStrength
 import de.gleex.pltcmd.model.signals.vision.Visibility
 import de.gleex.pltcmd.model.signals.vision.visibility
+import de.gleex.pltcmd.model.world.coordinate.CoordinateArea
 import kotlinx.coroutines.runBlocking
 import org.hexworks.amethyst.api.Consumed
 import org.hexworks.amethyst.api.Response
 import org.hexworks.amethyst.api.base.BaseFacet
+import org.hexworks.cobalt.datatypes.Maybe
 import org.hexworks.cobalt.logging.api.LoggerFactory
 
 /**
@@ -55,7 +62,8 @@ object Detects : BaseFacet<GameContext, DetectEntities>(
             val previousVisibility = lastSeen[seen] ?: Visibility.NONE
             logSeen(seeing, seen, visionStrength, previousVisibility)
             seeing.sighted(seen, visibility)
-            DetectedEntity(seen, visibility, previousVisibility, seeing, context)
+            val contact = seen.toContact(context)
+            DetectedEntity(contact, visibility, previousVisibility, seeing, context)
         } else {
             null
         }
@@ -76,4 +84,17 @@ object Detects : BaseFacet<GameContext, DetectEntities>(
         }
     }
 
+}
+
+fun PositionableEntity.toContact(context: GameContext): LocatedContact {
+    val location = CoordinateArea(currentPosition)
+    val area = context.world.areaOf(location)
+    val faction = asFactionEntity { it.reportedFaction.value }
+    // TODO
+    val corps: Maybe<Corps> = Maybe.empty()
+    val kind: Maybe<ElementKind> = Maybe.empty()
+    val rung: Maybe<Rung> = Maybe.empty()
+    val unitCount: Maybe<Int> = Maybe.empty()
+    val contact = Contact(faction, corps, kind, rung, unitCount)
+    return LocatedContact(area, contact)
 }
--- a/game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/facets/Detects.kt
+++ b/game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/facets/Detects.kt
@@ -29,7 +29,7 @@ object Detects : BaseFacet<GameContext, DetectEntities>(
 
     override suspend fun receive(message: DetectEntities): Response {
         val (visibleEntities, seeing, context) = message
-        val lastSeen: Map<PositionableEntity, Visibility> = seeing.forgetAll()
+        val lastSeen: Map<PositionableEntity, Visibility> = seeing.resetVision()
         visibleEntities
             .mapNotNull { seen -> createDetectedCommand(seen, seeing, lastSeen, context) }
             .apply {
@@ -54,7 +54,7 @@ object Detects : BaseFacet<GameContext, DetectEntities>(
         return if (visibility != Visibility.NONE) {
             val previousVisibility = lastSeen[seen] ?: Visibility.NONE
             logSeen(seeing, seen, visionStrength, previousVisibility)
-            seeing.rememberContact(seen, visibility)
+            seeing.sighted(seen, visibility)
             DetectedEntity(seen, visibility, previousVisibility, seeing, context)
         } else {
             null
--- a/game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/facets/Detects.kt
+++ b/game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/facets/Detects.kt
@@ -60,6 +60,7 @@ object Detects : BaseFacet<GameContext, DetectEntities>(
             logSeen(seeing, seen, visionStrength, previousVisibility)
             seeing.sighted(seen, visibility)
             val contact = seen.toContact(visibility, context)
+            seeing.asRememberingEntity { it.addContact(contact) }
             DetectedEntity(contact, visibility, previousVisibility, seeing, context)
         } else {
             null
--- a/game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/facets/Detects.kt
+++ b/game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/facets/Detects.kt
@@ -33,9 +33,9 @@ object Detects : BaseFacet<GameContext, DetectEntities>(
 
     override suspend fun receive(message: DetectEntities): Response {
         val (visibleEntities, seeing, context) = message
-        val lastSeen: Map<PositionableEntity, Visibility> = seeing.resetVision()
+        seeing.resetVision()
         visibleEntities
-            .mapNotNull { seen -> createDetectedCommand(seen, seeing, lastSeen, context) }
+            .mapNotNull { seen -> createDetectedCommand(seen, seeing, context) }
             .apply {
                 runBlocking {
                     forEach {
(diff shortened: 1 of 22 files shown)


Loomie:
Currently contacts are rememberd for their exact position. I prepared to check them in an area, but am not exactly sure how to do it. I think they should be assumed in an area around their current position depending on their (guessed) movement speed.

util/collections/src/main/kotlin/de/gleex/pltcmd/util/collections/CollectionUtils.kt
```diff
@@ -0,0 +1,19 @@
+package de.gleex.pltcmd.util.collections
+
+
+/** Applies each filter one after another and checks if only a single element is found. When all filters are applied all matching elements are given. */
+fun <T> Collection<T>.filterUntilFound(vararg filters: (T) -> Boolean): Set<T> {
```
Baret, line 5: Why does this return a `Set` when it always only contains up to one element? The implementation rather looks like a `filterSingleOrNull()` kind of method. Or `findFirstOrNull()` as the remaining filters are decarded as soon a single element is found.

util/knowledge/src/main/kotlin/de/gleex/pltcmd/util/knowledge/KnownByGrade.kt
```diff
@@ -0,0 +1,82 @@
...
+    fun reveal(newGrade: KnowledgeGrade) {
+        if (newGrade > grade) {
+            grade = newGrade
+        }
+    }
+
+    /**
+     * Returns the value of the [accessor] depending on the [revealed] value of this [Known]
+     */
+    fun <R> revealAt(minAmount: KnowledgeGrade, accessor: (T) -> R): Maybe<R> {
```
Baret, line 34: I really like this! Nice to use in implementations 👍

model/world/src/main/kotlin/de/gleex/pltcmd/model/world/coordinate/CoordinateArea.kt
```diff
@@ -8,6 +8,7 @@ import java.util.*
  */
 open class CoordinateArea(coordinateProvider: () -> SortedSet<Coordinate>) : Iterable<Coordinate> {
     constructor(coordinates: SortedSet<Coordinate>) : this({ coordinates })
+    constructor(coordinate: Coordinate) : this({ TreeSet<Coordinate>().apply { add(coordinate) } })
```
Baret, line 11: Wasn't there a `sortedSetOf(...)`? Relying directly on Java classes always looks strange ;)

Baret (commented):
First part of my review, more to come...

util/knowledge/src/main/kotlin/de/gleex/pltcmd/util/knowledge/KnownByGrade.kt
```diff
@@ -0,0 +1,82 @@
...
+        }
+        return false
+    }
+
+}
+
+/**
+ * Creates a [KnownByGrade] from this [Any] that is not revealed.
+ */
+fun <T : Any> T.nothingKnown() = KnownByGrade<T, KnownByGrade<T, *>>(
```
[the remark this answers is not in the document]
Loomie: Because the types are just aliases this works without a problem (see KnownTerrainTest):
```kotlin
val unknownByExtension: KnownTerrain = tile.nothingKnown()
```
It really is a `KnownByGrade` which provides itself as `SELF`.

util/knowledge/src/main/kotlin/de/gleex/pltcmd/util/knowledge/KnownByGrade.kt
Loomie: Maybe we should rename it to `revealAbove()`?

model/world/src/main/kotlin/de/gleex/pltcmd/model/world/coordinate/CoordinateArea.kt
Loomie: Yes, it is. But it lives in SetsJVM.kt and states:
> Returns a new java.util.SortedSet with the given elements.

So it is still bound to the JVM 🤷

util/knowledge/src/main/kotlin/de/gleex/pltcmd/util/knowledge/KnownByGrade.kt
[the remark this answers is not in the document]
Baret: The reason why you do not see the problem I am trying to point out here, yet, is that we do not yet use `mergeWith()`. But that method is the whole reason for having that `SELF` parameter. This snippet:
```kotlin
val some: KnownTerrain = KnownTerrain(WorldTile(coordinate, defaultTerrain), KnowledgeGrade.NONE)
val other: KnownTerrain = KnownTerrain(WorldTile(coordinate, defaultTerrain), KnowledgeGrade.FULL)
some.mergeWith(other)
```

Gives the error 
```
Type mismatch.
Required: Nothing
Found: KnownByGrade<WorldTile, *>
```
And the signature of mergeWith in the autocompletion is `mergeWith(other: Nothing)` because the "type" of SELF is always "*" in `KnownByGrade`.

Think ahead young padawan ;)
There should be one implementation per known type (aka type of origin). It needs to be an actual type, not a typealias. This is just not the use case for [a typealias](https://kotlinlang.org/docs/type-aliases.html).

util/knowledge/src/main/kotlin/de/gleex/pltcmd/util/knowledge/KnownByGrade.kt
Baret: Sounds like an off by one error by design. When I want something to be revealed at `FULL` I would always need to guess the next lower one because "above" means ">". I'd say `revealAt` is fine.

model/world/src/main/kotlin/de/gleex/pltcmd/model/world/coordinate/CoordinateArea.kt
Baret: Of course, because we use Kotlin/JVM here. But the actual implementation and where it comes from does not concern us. Maybe at some time kotlin comes with its own sortedSet implementation that is more efficient or whatever. In my eyes this is like one should prefer using interfaces instead of actual implementations.
And probably also [this](https://kotlinlang.org/docs/coding-conventions.html#immutability).

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/facets/Detects.kt
```diff
@@ -64,16 +66,29 @@ object Detects : BaseFacet<GameContext, DetectEntities>(
...
-        previousVisibility: Visibility
+        visibility: SignalStrength
     ) {
         if (log.isDebugEnabled()) {
             val who = seeing.logIdentifier.padEnd(12)
             val what = seen.logIdentifier.padEnd(12)
             val where = seen.currentPosition.toString().padEnd(12)
             val how = visibility.asRatio()
-            log.debug("$who sees $what at $where with signal strength $how (last visibility was $previousVisibility)")
+            log.debug("$who sees $what at $where with signal strength $how")
```
Baret, line 76: `visiblility` is actually a `Visibility`. It is semantically more accurate and would also more oviously lead to a more descriptive log message:
```suggestion
            log.debug("$who sees $what at $where with visibility $how")
```

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/facets/Detects.kt
```diff
@@ -64,16 +66,29 @@ object Detects : BaseFacet<GameContext, DetectEntities>(
...
 
 }
+
+fun ElementEntity.toContact(visibility: Visibility, context: GameContext): Contact {
+    // TODO remember where the contact was spotted
+//    val location = CoordinateArea(currentPosition)
+//    val area = context.world.areaOf(location)
+    return Contact(
+        origin = this,
+        when (visibility) {
```
Baret, line 88: Isn't there a `Visibilty.asRatio()` and a `KnowledgeGrade.fromRatio()` or something like that? Thi when breaks when we have a different number of values for visibility and grade (or is it already?).

Baret (changes requested):
9 more unviewed files... But you can already work an those comments ;)

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/attributes/memory/elements/Contact.kt
```diff
@@ -0,0 +1,52 @@
...
+import org.hexworks.cobalt.datatypes.Maybe
+
+/**
+ * Holds information about an [Element] that was reported (directly or indirectly, by vision, sound or assumption).
+ * It may not be accurate and mostly incomplete!
+ */
+typealias Contact = KnownByGrade<ElementEntity, *>
+
+// basic information is always available
+val Contact.corps: Maybe<Corps>
```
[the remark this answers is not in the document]
Loomie: It's just because the `origin` is an `ElementEntry` I added all accessor of it. Because I don't have a model class 👅

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/attributes/memory/elements/Contact.kt
```diff
@@ -0,0 +1,52 @@
...
+    get() = Maybe.of(origin.element.corps)
+val Contact.kind: Maybe<ElementKind>
+    get() = Maybe.of(origin.element.kind)
+
+// details of the element depend on how much we know about it
+val Contact.rung: Maybe<Rung>
+    get() = revealAt(KnowledgeGrade.LOW) { it.element.rung }
+val Contact.faction: Maybe<Faction>
```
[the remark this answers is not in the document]
Loomie: The affiliation is relative to another faction.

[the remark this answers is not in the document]
Baret: Yes. And...?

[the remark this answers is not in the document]
Loomie: Reworked it: b337c20d

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/facets/Detects.kt
```diff
@@ -45,17 +47,17 @@ object Detects : BaseFacet<GameContext, DetectEntities>(
     private fun createDetectedCommand(
         seen: PositionableEntity,
         seeing: SeeingEntity,
-        lastSeen: Map<PositionableEntity, Visibility>,
         context: GameContext
     ): DetectedEntity? {
         val seenPosition = seen.currentPosition
         val visionStrength = seeing.vision.at(seenPosition)
```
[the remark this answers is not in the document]
Loomie: I think it is fine so. The LookAround looks in a direction and says: I see this part of the world. `Detects` then checks if in that part of the world something is interesting. But as it is the glue between the both it could also be put at the other side. I like the current small data that is transfered. The type is just an `EntitySet<Positionable>` rather than a `Map<Positionable, SignalStrength>`. Although I like the early filtering and thus less `Message` sending/receiving.

But the  Vision has nothing to do with the knowledge of this issue/pull request ;)

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/facets/Detects.kt
Loomie: > `visiblility` is actually a `Visibility`. 

Nope, it is a `SignalStrength`. For debugging the actual value rather then the category (GOOD/POOR/NONE) is more interesting.

game/engine/src/main/kotlin/de/gleex/pltcmd/game/engine/systems/facets/Detects.kt
Loomie: ℹ️ This is now in `ReportContacts`.

> when we have a different number of values for visibility and grade (or is it already?).

It is already different: 3 Visiblities and 5 KnowledgeGrades.

> Isn't there a `Visibilty.asRatio()`

No.

I think for knowledge we dont need the full `SignalStrength.asRatio()` as it is not so fine grained.

Loomie:
Thanks for the feedback. I changed the `Contact` class once again.

util/knowledge/src/main/kotlin/de/gleex/pltcmd/util/knowledge/KnownByGrade.kt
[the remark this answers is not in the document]
Loomie: KnowByGrade.mergeWith() works. The KnownByGradeTest ensures this. So if something does not work it is just the generic typing.

This actually works: `knownByExtension.mergeWith(unknownByAlias)`. So if the type is inherited instead of using the typealias it is fine. And for that type the extension functions also apply. The alias is a simplification which increases readablility in some cases. But as Kotlin often does not require to declare the type it is even more readable without the type declaration. Like in `val unknownByExtension = tile.unknown()`. I added tests for this in 24a7fc7f.

On the other hand it is strange that Kotlin can use the typealias also for the constructor call although it is not fully typed 🤨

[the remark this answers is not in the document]
Loomie: Oh, if you really want an own class you/we could use composition over inheritance. And that also allows the usage of a data class that has the known object as field:
```kotlin
data class KnowTerrainData(val known: KnownByGrade<WorldTile, *>)
```
With this construct we can keep the generated equals :) That was not possible if the parent is a data class. And the access depends on the user: If the code wants to read the knowledge is used `knownTerrain.known` to access the `Known` interface. If business logic wants to access terrain it can directly use the methods at KnowTerrainData. But that is almost what I achived with externsion functions and typealias.

PR #140, opened 2021-03-27
 map persistence resolves #96

### LINKED ISSUE #96: Save&Load the WorldMap
**Basic idea/big picture**
When developing you have to wait for the map generation every single time. It would be great if the map generated with the `DEBUG_MAP_SEED` was in a saved file in the resources folder so you can skip the generation process by simply loading it.
In the long run this feature may also be useful as the basis for savegames.

**Details**
We already have an option to quickly toggle the intro and if we want to use the debug map seed or generate a random world. I would expand this to a "dev mode" which can be enabled/disabled in the `GameOptions`. Dev mode makes the game startup quicker and sets you in a specific state that we will most probably need throughout development (like a fixed map where you know in which locations you can test stuff). So for now it would skip the intro and load the map from the file system.

The ToDos of this issue are as follows (open for discussion/change, of course):
- Create a Keybinding to save the current map to a file while in the `GameView`
  - Preferably one of the F-Keys and not F1 as it will probably be the "Help" keybinding at some point
  - The filename should not overwrite an existing file that contains a different map. So it should contain the map seed (to easily reproduce it if needed) and probably a timestamp.
- On startup load a file that is committed in `src/main/resources` when `GameOption.devMode` == `true`
- Create a Unittest that checks if the map generation with the debug seed results in the same map that is stored in `src/main/resources`
  - It should test the saving and loading (compare the newly created file with the checked in one and compare the loaded map with the generated one)
  - This helps us to recognize changes in world generation. When something changes this test should become red so we know that we need to "export" the "debug map" again and commit it.

**Technical background**
I am not yet committed on the file format. Maybe something human readable is nice but on the other hand a list of 25 million tiles in text is not quite maintanable by hand :D
First I thought of JSON but it might be too hard on the file size. Just use whatever is handy. Maybe have a look for a nice kotlin lib.

Maybe we can also create our own file format that creates a zip with every sector in a single file. It might make saving changes easier for whatever reason one might use that ^^
Or a simple height- and terrain map in one file each zipped togeter:
```
debug_map.zip (oder .pcm = pltcmd map)
  debug_map.hm (hm = heightmap)
  debug_map.tm (tm = terrainmap)
```
with the heightmap being:
```
34567
35566
44456
55566
66677
```
and the terrainmap being
```
FFFGG
FFGGH
FGMMM
HHHGG
HHFGM
```
Of course a mix of both ideas is also possible.

**Related**
Direct follow up issue: #107 😃

Loomie:


Loomie:
Using ProtoBuf as format because it is binary and pretty small. Although it is expirmental.

Loomie:
After generating a map it is saved to a file. When the game is started again it loads the map from the file.

Loomie:
not done:
> Create a Unittest that checks if the map generation with the debug seed results in the same map that is stored

Sounds more like an integration test that checks the interoperability between generation and storage.

Loomie:
I added a menu to choose if the last map should be loaded or a new one generated.

game/options/src/main/kotlin/de/gleex/pltcmd/game/options/GameOptions.kt
```diff
@@ -45,6 +45,7 @@ object GameOptions {
             } else {
                 System.currentTimeMillis()
             }
+    val MAP_FILE = if (USE_DEBUG_MAP_SEED) "develop" else "last"
```
Baret, line 48: I wanted to extend this boolean to something like "dev mode" anyways. Maybe you could introduce it now.
Then we have one boolean we derive different things of like special UI elements, skipping intro, loading the dev map etc.
So for now it would be used here in this line and I'd probably simply set `USE_DEBUG_MAP_SEED = DEV_MODE`.

game/serialization/pom.xml
```diff
@@ -0,0 +1,25 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>game</artifactId>
+        <groupId>de.gleex.pltcmd.game</groupId>
+        <version>0.2.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>serialization</artifactId>
```
Baret, line 12: Name and description, please.

game/serialization/src/main/kotlin/de/gleex/pltcmd/game/serialization/Storage.kt
```diff
@@ -0,0 +1,45 @@
...
+ * Manages persistent data.
+ */
+@OptIn(ExperimentalSerializationApi::class)
+internal object Storage {
+    private val log = LoggerFactory.getLogger(Storage::class)
+    private val dataFolder = File("data")
+
+    init {
+        if (!dataFolder.exists() && !dataFolder.mkdirs()) {
+            log.error("Failed to create folder ${dataFolder.absolutePath}")
```
Baret, line 20: And then...? Proceed with this error? Dead programs tell no lies. Either throw an exception and let it crash or make `Stoage` some kind of inactive so that you don't run into "some random IOException" when you try to save/load something.

game/serialization/src/main/kotlin/de/gleex/pltcmd/game/serialization/world/MapStorage.kt
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Stores and loads map data.
+ */
+@OptIn(ExperimentalTime::class)
+object MapStorage {
+    private val log = LoggerFactory.getLogger(MapStorage::class)
+    private const val fileSuffix = ".map"
+
+    /** Save the given map to the specified file. */
+    fun save(map: WorldMap, fileId: String) {
```
Baret, line 20: What is the `fileId`? Docs plz

game/serialization/src/main/kotlin/de/gleex/pltcmd/game/serialization/world/WorldMapDao.kt
```diff
@@ -0,0 +1,28 @@
...
+import de.gleex.pltcmd.model.world.sectorOrigin
+import kotlinx.serialization.Serializable
+
+/** Stores/loads map data. */
+@Serializable
+data class WorldMapDao(val tiles: Collection<WorldTile>) {
+
+    /** Convert this data back to a model. */
+    fun toMap(): WorldMap {
+        val sectors = tiles
```
Baret, line 15: Don't we have those lines somwhere else already? Maybe we should have some kind of `Collection<WorldTile>.toSectorSet()`.

Baret (commented):
This looks good all in all. There are just some litte improvements to make.

I really like the menu! Opens up possibilities for the future! And the simplistic design is totally cool for me :)

game/application/src/main/kotlin/de/gleex/pltcmd/game/application/main.kt
```diff
@@ -53,9 +55,41 @@ open class Main {
...
         showTitle(screen, tileGrid)
+        selectMap(screen, tileGrid)
+    }
 
-        generateMap(screen, tileGrid) { generatedMap ->
-            runGame(generatedMap, screen, tileGrid)
+    protected open fun selectMap(screen: Screen, tileGrid: TileGrid) {
+        val mapFile = GameOptions.MAP_FILE
+        val loadedMap = MapStorage.load(mapFile)
+        if (loadedMap == null && UiOptions.SKIP_INTRO.not()) {
```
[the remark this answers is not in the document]
Loomie: Because loading the map takes so long that in the meantime you see the intro. That's why the map is loaded first and the title screen is shown afterwards. But I will extend the comment.

game/serialization/src/main/kotlin/de/gleex/pltcmd/game/serialization/Storage.kt
```diff
@@ -0,0 +1,45 @@
...
+            log.warn("Cannot load data from $file because it is not readable!")
+            return null
+        }
+        val bytes = file.readBytes()
+        return ProtoBuf.decodeFromByteArray<R>(bytes)
+    }
+}
+
+/** Remove path navigation from name like "../../root" and an extension */
+val String.fileBasename: String
```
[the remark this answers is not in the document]
Loomie: I also thought about it during development. It was a File but that was unnecessary complex. And it was a function which is just not Kotlin style ;)

game/serialization/src/main/kotlin/de/gleex/pltcmd/game/serialization/world/MapStorage.kt
```diff
@@ -0,0 +1,42 @@
...
+        val fileName = toMapFile(fileId)
+        val (map, duration) = measureTimedValue {
+            val dao = Storage.load<WorldMapDao>(fileName)
+            dao?.toMap()
+        }
+        log.info("loaded $map from $fileName in $duration")
+        return map
+    }
+
+    private fun toMapFile(fileId: String) = fileId.fileBasename + fileSuffix
```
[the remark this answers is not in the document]
Loomie: 👍

game/serialization/src/main/kotlin/de/gleex/pltcmd/game/serialization/world/WorldMapDao.kt
Loomie: No, we have the `Sector` class for the model. But that is skipped in the persistence. It is helpful for accessing the world, but not for storage.

game/serialization/src/main/kotlin/de/gleex/pltcmd/game/serialization/Storage.kt
```diff
@@ -0,0 +1,47 @@
...
+        }
+        val bytes = file.readBytes()
+        return ProtoBuf.decodeFromByteArray<R>(bytes)
+    }
+
+    private val StorageId.file
+        get() = File(dataFolder, sanitizeFilename(id) + "." + type)
+}
+
+/** Remove path navigation from name like "../../root" and an extension */
```
Baret, line 46: Would be nice to know what exectly is being returned. when I pass `/tmp/../myFile.txt`, do I get `/tmp/myFile.txt`? or `myFile`? Maybe you could add such an example to the kdoc (and the... unittest? ;) )

Loomie: I added a UnitTest to document the result for some (special) inputs

pom.xml
```diff
@@ -30,9 +30,10 @@
 		<maven.compiler.target>1.8</maven.compiler.target>
 
 		<kotlin.compiler.jvmTarget>1.8</kotlin.compiler.jvmTarget>
-		<kotlin.version>1.4.20</kotlin.version>
+		<kotlin.version>1.4.30</kotlin.version>
```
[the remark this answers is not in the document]
Loomie: `kotlinx.serialization` 1.1.0 release said:
> uses release version of Kotlin 1.4.30.

That's why I updated the Kotlin version.

model/world/src/main/kotlin/de/gleex/pltcmd/model/world/terrain/Terrain.kt
```diff
@@ -1,7 +1,9 @@
 package de.gleex.pltcmd.model.world.terrain
 
+import kotlinx.serialization.Serializable
 import kotlin.random.Random
 
+@Serializable
 data class Terrain private constructor(val type: TerrainType, val height: TerrainHeight) {
```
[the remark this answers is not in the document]
Loomie: I tried it now: Adding an element in `TerrainHeight` inbetween existing elements:
```kotlin
    ONE(1),
    ELEVEN(11),
    TWO(2),
    ...
```
After loading a previous map all TWO became ELEVEN. As expected I would say. Appending new elements to the end of an enum will be fine.

Loomie:
Have another look.

model/world/src/main/kotlin/de/gleex/pltcmd/model/world/terrain/Terrain.kt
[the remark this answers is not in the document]
Baret: Ah so it seems to save the ordinal.
That's what I'd expected. Adding new values would be ok, changing existing ones or removing is bad. 

For the terrain enums I'd only expect new values. Maybe we need more heights. And we might add 1 or 2 additinal terrain types (I am thinking about `agricultural` for example).

pom.xml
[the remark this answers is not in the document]
Baret: And 1.4.32 recommends `serialization` 1.1.0 :D

But I just see directly updated to kotlin 1.5.10... well, cool! We'd have done that anyways at some point.

Loomie:
The main menu now lists all stored maps with their name (which is still hard coded).

game/serialization/src/main/kotlin/de/gleex/pltcmd/game/serialization/Storage.kt
```diff
@@ -0,0 +1,82 @@
...
+        if (!file.canRead()) {
+            log.warn("Cannot load data from $file because it is not readable!")
+            return null
+        }
+        val bytes = file.readBytes()
+        return ProtoBuf.decodeFromByteArray<R>(bytes)
+    }
+
+    /** lists all stored entries of the given type */
+    inline fun listAll(type: String): List<StorageId> {
```
Baret, line 43: Why is this an inline fun?

PR #142, opened 2021-04-11
Issue 141 networking basics

Loomie:


Loomie:
Using https://ktor.io/ for networking.

Loomie:
The log area in the `GameView` now gets its messages by network from the server. The server listenes for events on the EventBus, extracts relevant data, transmits it by network and creates a new UI event. The UI listens for those events and displays the messages.

For serialization I used kotlinx.serialization with Protocol Buffers as in #140

Issue #144, opened 2021-05-04
UI displays the known world of an entity


Baret:
**Basic idea/big picture**
After #143 entities now have knowledge about the terrain they have already seen. This should also be reflected in the UI.

**Details**
The UI should display the `KnownWorld` of one entity, usually the HQ, of course. It would be a cool debug feature to swap the entity at runtime but that's just a nice to have.

**Technical background**
Each block displays one coordinate of the `KnownWorld`. It is either unknown or known aka revealed. The `TileRepository` can already serve tiles for unknown (actually "ungenerated" terrain), so we can use that for a fist step.
As terrain can currently only be revealed once and the revealed state is always "the truth" it is enough for each block to only receive one "your coordinate has been revealed"-event.
I had a first try of this in #143 and failed. But after thinking about it there are two ways I could imagine. Either the `KnownWorld` offers an own way of adding listeners for specific coordinates. Something like `notifyOnce(coordinate: Coordinate, callback: (Coordinate) -> Unit)` that holds callbacks for each coordinate in a map and removes them after they have been called.
Or `KnownByBoolean.revealed` becomes an `ObservableValue<Boolean>` that each block can listen to.
In either case the subscription should be discarded after one event to not keep millions of references over the lifetime of the application.

If necessary I have one possible optimization in mind: Only the blocks in the visible part of the `GameView` need to set their terrain tile.

PR #150, opened 2021-05-22
Feature #144 display known world

Loomie:
Currently only the terrain is hidden, but the elements are still visible. It is more a proof of concept than a full solution ;)


Baret:
I'd say thats perfectly fine for now. Nothing more than what I aimed at with this issue :)
I will take a look when I find some spare time

PR #151, opened 2021-07-18
Resolves #100 created module pathfinding

### LINKED ISSUE #100: Create module Pathfinding
**Basic idea/big picture**
We will need different "path finders" for elements to plan their way through the world. This best fits into an own module which is best located under `util`. Maybe it is too dependent on `model.world` so we might also need to put it somewhere else.

The idea is to have one interface `Pathfinder` that has different implementations and that can be used to calculate a path from a start location to a destination.

**Details**
Currently we only have a "straight line" "pathfinding" but we can acutally make use of that ingame (flying units) so it is probably a good idea to start creating a `StraightLinePathfinder`. Later we can create more complex (and "intelligent") ones using Dijkstramaps or A*-Pathfinding, which will be looked into in #82

With this issue we should also fix this TODO:
https://github.com/Baret/pltcmd/blob/5db84c66e7e6e7c5160f5857f55ac5386e502aff/model/mapgeneration/pom.xml#L14

**Technical background**
The interface exposed by the module and used in `game` should be as simple as possible. I would like to rather have lots of different implementations than to have like 2 or 3 that are confusingly configurable. For example you might have "the one A*-pathfinder". When you want to get somewhere stealthily/with much cover you tell it to use this value for forest terrain and this value for grasslands terrain and so on. When you want the quickest path you need to provide a different set of configuration/parameters.

This does not sound pretty in my ears. I would rather like to simply use `PreferCoverPathfinder` or `FastestInfantryPathfinder` or `FastestTrackedVehiclePathfinder` and call `findPath(currentLocation, destination)` on it and that's it.

This issue might be a duplicate or at least tightly related to #82

Loomie:


Baret (approved):
Looks good!
Maybe we later need some factoty for pathfinders instead of using concrete implementations directly. But this is a solid base!

PR #163, opened 2021-10-31
Resolves #145 entities merge their knowledge into comrades

### LINKED ISSUE #145: Entities merge their knowledge into the HQ
**Basic idea/big picture**
In the final game there will be stationed elements and deployed elements. When a deployed element gets back to the main base and transfers to stationed status, it will give all its knowledge to the HQ (their knowledge will be merged). As I see this pretty far in the future but want to see it in action much earlier I have an idea of using it already.
Entities can simply "drop" their knowledge when they get back to the HQ. So as soon as they are on the same tile as the HQ, they merge their knowledge into the HQ's knowledge. This makes it possible to send out elements to scout areas and reveal them as soon as they RTB (which would be a useful radio command for this btw).

**Details**
Actually this is pretty simple. As soon as an element moves onto the same field as the HQ they merge their knowledge. Possibly in both ways so that elements learn something, too.
We could also try to let entities leran from each other. So we might not check for "at HQ location" but for "at friendly element location". But I don't think that makes much sense.

This needs #143 and #144 to be done, so you can actually see the updates you get from your elements.

Loomie:
If two or more entities of the same faction are on the same tile, the updated entity sends its knowledge to the others.


Baret (approved):
Looks good! Nice and simple :)
The two comments are just stuff we gotta remember for the future.
I'm gonna merge it!

